PR #8, opened 2018-06-13
Database model setup

luinnarn:


src/main/java/eu/execom/hawaii/model/PublicHoliday.java
```diff
@@ -0,0 +1,31 @@
...
+  @ManyToOne
+  private LeaveProfile leaveProfile;
+
+  @NotNull
+  private String name;
+
+  @NotNull
+  private LocalDate date;
+
+  private boolean deleted;
```
stanislavgrujic: Why not make that one `active` as well, to have it consistent throughout the code base?

src/main/java/eu/execom/hawaii/model/absence/Absence.java
```diff
@@ -0,0 +1,31 @@
...
+public abstract class Absence extends BaseEntity implements Serializable {
+
+  private static final long serialVersionUID = -3525779850487542671L;
+
+  @NotNull
+  private String name;
+
+  private String comment;
+
+  private boolean deleted;
```
stanislavgrujic: Think if it makes sense to use either naming `deleted` for such fields, or `active`. I'm fine with both, but use it consistently throughout the code base.

src/main/resources/application.properties
```diff
@@ -1,7 +1,12 @@
...
+spring.datasource.url=jdbc:mysql://localhost:3306/hawaii?useSSL=false
 spring.datasource.username=root
 spring.datasource.password=root
 
-################### Hibernate Configuration ##########################
-spring.jpa.hibernate.ddl-auto=create
\ No newline at end of file
+################### Hibernate/Spring JPA Configuration ###############
+spring.jpa.properties.hibernate.dialect = org.hibernate.dialect.MySQL5InnoDBDialect
+spring.jpa.hibernate.ddl-auto=update
```
stanislavgrujic: I think you want liquibase to do the database update, and here set it to `none`.

mvnw.cmd
```diff
@@ -41,7 +41,7 @@
 @REM set %HOME% to equivalent of $HOME
 if "%HOME%" == "" (set "HOME=%HOMEDRIVE%%HOMEPATH%")
 
-@REM Execute a user defined script before this one
+@REM Execute a userDao defined script before this one
```
stanislavgrujic: This change does not make much sense. Please revert.

src/main/resources/liquibase.properties
```diff
@@ -0,0 +1,7 @@
+#MySql
+driver=com.mysql.jdbc.Driver
+url=jdbc:mysql://localhost:3306/hawaii?useSSL=false
+username=root
+password=root
+changeLogFile=src/main/resources/db/changelog/changelog-master.xml
+verbose=true
```
stanislavgrujic: Is it possible to avoid duplicate database properties configuration in this properties file and in the `application.properties`?

stanislavgrujic:
General comment - it would be nice if you have the same ordering of fields in the PUML diagram and in the actual classes - it will make it much easier to compare them. Please update it.

src/main/resources/db/changelog/changelog-master.xml
```diff
@@ -0,0 +1,11 @@
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+  <property name="initialId" value="initial_data" dbms="mysql"/>
+  <property name="migrationId" value="migration" dbms="mysql"/>
```
stanislavgrujic: Please put a more meaningful name of the change, rather than `migration`.

src/main/resources/liquibase.properties
dujfalusi: Liquibase maven plugin inside pom.xml have configuration part:
```
<configuration>
    <propertyFile>src/main/resources/liquibase.properties</propertyFile>
</configuration>
```
Where by default takes all properties.

It could be possible to configure from application.properties, something like: 

```
 <configuration>
          <driver>${env.my.sql.driver}</driver>
          <url>${env.spring.datasource.url}</url>
          <username>${env.spring.datasource.username}</username>
          <password>${env.spring.datasource.password}</password>
          <propertyFile>src/main/resources/liquibase.properties</propertyFile>
</configuration>
```

OR moving all properties from liquibase.properties to application.properties with an environment variables like:
```
spring.liquibase.password=${DB_PASSWORD}
```

src/main/resources/db/changelog/changelog-master.xml
luinnarn: Something like `schema_update`?

stanislavgrujic: Ah, sorry. My bad, now I see it's usage. Please, disregard this comment.

src/main/resources/liquibase.properties
stanislavgrujic: Is there a question? Which one would you use?

dujfalusi: I suggest to stay like this. (for now)

I am just tried to implement both options: 
1. add a separate database.properties configured with filters and resources inside pom.xml/build and failed to bind.
2. this spring liquibase properties in application properties 
```spring.liquibase.password=...``` 
spring boot 2.02 doesn't recognize as liquibase configuration.

stanislavgrujic: OK.

PR #9, opened 2018-06-14
B feature/user management

luinnarn:
Issue #13


src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,23 @@
...
+
+@Slf4j
+@ControllerAdvice
+public class GlobalExceptionHandler {
+
+  @ResponseStatus(HttpStatus.NOT_FOUND) // 404
+  @ExceptionHandler(EntityNotFoundException.class)
+  public void handleEntityNotFoundException(HttpServletRequest request, Exception exception) {
+    log.error("Requesting resource: {} raised exception: {}", request.getRequestURI(), exception);
+  }
```
stanislavgrujic: Please also add handling for all other exceptions, which will return HTTP status code 500.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -0,0 +1,39 @@
...
+    return userRepository.findAll();
+  }
+
+  public User getUser(Long id) {
+    return userRepository.getOne(id);
+  }
+
+  public User getUserByEmail(String email) {
+    User user = userRepository.findByEmail(email);
+    if(user == null) {
```
stanislavgrujic: This is still not formatted properly.

PR #11, opened 2018-06-14
F feature/sidebar component

aberonja:
Created sidebar component with temporary style


src/main/frontend/src/components/StyledLink.js
```diff
@@ -1,11 +1,20 @@
 import styled from 'styled-components';
-import { prop } from 'styled-tools';
+import { prop, ifProp } from 'styled-tools';
 
 export const StyledLink = styled.span`
```
iscekic: I dislike the name of this component, let's hear some suggestions. 😄

dusanfle: NavigationLink, maybe?

PR #21, opened 2018-06-28
B feature oauth2

luinnarn:
Issue #16: basic Google login functionality. At this point, the only check is whether the signed in user is from the execom.eu domain.


src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -27,6 +28,18 @@ public void handleEntityExistsException(HttpServletRequest request, Exception ex
     logException(request, exception);
   }
 
+  @ResponseStatus(HttpStatus.CONFLICT) // 409
+  @ExceptionHandler(DataIntegrityViolationException.class)
```
stanislavgrujic: Is this exception thrown only if a unique key constraint is violated, or there may be some other cases when this is also thrown (e.g., a relation between two tables is not satisfied)?

src/main/java/eu/execom/hawaii/api/controller/SignInController.java
```diff
@@ -0,0 +1,75 @@
...
+    if (!email.endsWith(EXECOM_DOMAIN)) {
+      return new ResponseEntity<>(HttpStatus.UNAUTHORIZED);
+    }
+
+    User user = userService.getByEmail(email);
+
+    final Collection<GrantedAuthority> authorities = new ArrayList<>();
+    authorities.add(new SimpleGrantedAuthority(user.getUserRole().name()));
+
+    final Authentication authentication = new PreAuthenticatedAuthenticationToken(user, null, authorities);
```
stanislavgrujic: Please remove `final` keywords, we do not use it in our code base.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -31,13 +32,26 @@ public UserService(UserRepository userRepository) {
...
   }
 
+  /**
+   * Retrieves a User with a specific id
+   *
+   * @param id User email
+   * @return User with specified email if exists
+   * @throws EntityNotFoundException if a user with given id is not found
+   */
+  @Transactional
```
stanislavgrujic: Please add `readOnly = true`.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -31,13 +32,26 @@ public UserService(UserRepository userRepository) {
...
+  public User getUserById(Long id) {
+    return userRepository.getOne(id);
```
stanislavgrujic: Please add `readOnly = true`.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -31,13 +32,26 @@ public UserService(UserRepository userRepository) {
...
+  }
+
   /**
    * Retrieves a User with a specific email.
    *
    * @param email User email
    * @return User with specified email if exists
    * @throws EntityNotFoundException if a user with given email is not found
    */
+  @Transactional
```
stanislavgrujic: Please add `readOnly = true`.

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
luinnarn: This is also thrown in other cases such as when a foreign key constraint or a NotNull constraint is not satisfied.

stanislavgrujic: Knowing that and reading this https://httpstatuses.com/409, does that behavior fit into description?

luinnarn: 409 should be thrown whenever a resource conflict would be caused by fulfilling the request. Doesn't a foreign key or not null constraint represent such a resource conflict?

stanislavgrujic: 4xx are client errors. Client cannot affect the foreign key, and null values should be validated upon receiving the client request, not on the database error.

PR #22, opened 2018-07-03
B feature/team management

luinnarn:
Issue #14


src/main/java/eu/execom/hawaii/api/controller/TeamController.java
```diff
@@ -0,0 +1,72 @@
...
+public class TeamController {
+
+  private TeamService teamService;
+
+  @Autowired
+  public TeamController(TeamService teamService) {
+    this.teamService = teamService;
+  }
+
+  private static final ModelMapper MAPPER = new ModelMapper();
```
stanislavgrujic: Please move static fields at the top of the class.

Issue #23, opened 2018-07-05
Leave profile controller


luinnarn:
Create endpoints for leave profile management

PR #24, opened 2018-07-05
B feature/leave profile management

luinnarn:
Issue #23

### DIFF
--- a/src/main/java/eu/execom/hawaii/api/controller/LeaveProfileController.java
+++ b/src/main/java/eu/execom/hawaii/api/controller/LeaveProfileController.java
@@ -1,0 +1,72 @@
+package eu.execom.hawaii.api.controller;
+
+import eu.execom.hawaii.dto.LeaveProfileDto;
+import eu.execom.hawaii.model.LeaveProfile;
+import eu.execom.hawaii.service.LeaveProfileService;
+import org.modelmapper.ModelMapper;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.DeleteMapping;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+@RestController
+@RequestMapping("/leaveprofiles")
+public class LeaveProfileController {
+
+  private static final ModelMapper MAPPER = new ModelMapper();
+
+  private LeaveProfileService leaveProfileService;
+
+  @Autowired
+  public LeaveProfileController(LeaveProfileService leaveProfileService) {
+    this.leaveProfileService = leaveProfileService;
+  }
+
+  @GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<List<LeaveProfileDto>> getLeaveProfiles() {
+    var leaveProfiles = leaveProfileService.getAll();
+    var leaveProfileDtos = leaveProfiles.stream().map(LeaveProfileDto::new).collect(Collectors.toList());
+    return new ResponseEntity<>(leaveProfileDtos, HttpStatus.OK);
+  }
+
+  @GetMapping(value = "/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<LeaveProfileDto> getLeaveProfile(@PathVariable Long id) {
+    var leaveProfile = leaveProfileService.getById(id);
+    var leaveProfileDto = new LeaveProfileDto(leaveProfile);
+    return new ResponseEntity<>(leaveProfileDto, HttpStatus.OK);
+  }
+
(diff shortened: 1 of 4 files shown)


src/main/java/eu/execom/hawaii/api/controller/LeaveProfileController.java
```diff
@@ -0,0 +1,72 @@
...
+  public ResponseEntity<LeaveProfileDto> getLeaveProfile(@PathVariable Long id) {
+    var leaveProfile = leaveProfileService.getById(id);
+    var leaveProfileDto = new LeaveProfileDto(leaveProfile);
+    return new ResponseEntity<>(leaveProfileDto, HttpStatus.OK);
+  }
+
+  @PostMapping(consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<LeaveProfileDto> createLeaveProfile(@RequestBody LeaveProfileDto leaveProfileDto) {
+    var leaveProfile = MAPPER.map(leaveProfileDto, LeaveProfile.class);
+    leaveProfile = leaveProfileService.save(leaveProfile);
```
stanislavgrujic: Why do you need to reassign variable?

src/main/java/eu/execom/hawaii/api/controller/LeaveProfileController.java
```diff
@@ -0,0 +1,72 @@
...
+    var leaveProfileDtoResponse = new LeaveProfileDto(leaveProfile);
+    return new ResponseEntity<>(leaveProfileDtoResponse, HttpStatus.CREATED);
+  }
+
+  @PutMapping(consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<LeaveProfileDto> updateLeaveProfile(@RequestBody LeaveProfileDto leaveProfileDto) {
+    var leaveProfile = MAPPER.map(leaveProfileDto, LeaveProfile.class);
+    leaveProfile = leaveProfileService.save(leaveProfile);
```
stanislavgrujic: Same here, why do you need to reassign it?

src/main/java/eu/execom/hawaii/service/LeaveProfileService.java
```diff
@@ -0,0 +1,66 @@
...
+  public LeaveProfileService(LeaveProfileRepository leaveProfileRepository) {
+    this.leaveProfileRepository = leaveProfileRepository;
+  }
+
+  /**
+   * Retrieves a list of all leave profiles from repository.
+   *
+   * @return a list of all leave profiles
+   */
+  public List<LeaveProfile> getAll() {
```
stanislavgrujic: You are invoking `findAll` from the repository. Why name the wrapper function differently?

src/main/java/eu/execom/hawaii/service/LeaveProfileService.java
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Retrieves a leave profile with a specific id.
+   *
+   * @param id Leave profile id
+   * @return Leave profile with provided id if exists
+   * @throws EntityNotFoundException if a leave profile with given id is not found
+   */
+  public LeaveProfile getById(Long id) {
+    return leaveProfileRepository.findById(id).orElseThrow(EntityNotFoundException::new);
```
stanislavgrujic: You can achieve the same thing by using `leaveProfileRepository.getOne(id)`.
Make sure you know the API of the frameworks/libraries you are using.

src/main/java/eu/execom/hawaii/api/controller/LeaveProfileController.java
luinnarn: Because it gets a generated id assigned upon persisting to database, which should be contained in the response DTO.

src/main/java/eu/execom/hawaii/api/controller/LeaveProfileController.java
luinnarn: Following the reasoning from the comment above, this was left in case an update is attempted on an entity that's not in the database (in which case the repository `save()` method would persist it as a new entry). this should in fact never happen anyway, so this is unnecessary and will be removed.

src/main/java/eu/execom/hawaii/service/LeaveProfileService.java
luinnarn: Some service methods invoke repository `get` methods (`getOne()`), and some invoke repository `find` methods (`findAll()`). The idea was to make the naming of the service methods consistent among themselves.

stanislavgrujic: It's a good intention, but there is a reason why there are methods `find` and `get` and it's a convention known to most developers how each of them behaves. This way you are deviating from the convention, which makes it more difficult to grasp the code.

src/main/java/eu/execom/hawaii/api/controller/LeaveProfileController.java
stanislavgrujic: The object you passed is updated with an ID, save method does not create a new object, so you are assigning the object to itself. Try to debug the code to see that the instance is the same, and that it already has it's ID (or please let me know if I'm missing something).

luinnarn: Actually the repository method returns a new instance, this is from the CrudRepository.save() javadoc:
>Saves a given entity. Use the returned instance for further operations as the save operation 
>might have changed the entity instance completely.

stanislavgrujic: Cool. Thanks for the clarification, you can disregard my comment then. :)

src/main/java/eu/execom/hawaii/service/LeaveProfileService.java
luinnarn: Should this distinction be applied to the controller methods then as well?

PR #27, opened 2018-07-10
Added swagger

dujfalusi:
#26


src/main/java/eu/execom/hawaii/security/SecurityConfiguration.java
```diff
@@ -6,26 +6,26 @@
...
         .and()
-        .logout()
-        .invalidateHttpSession(true)
-        .deleteCookies("JSESSIONID")
-        .logoutSuccessUrl("http://localhost:3000/")
-        .permitAll();
+          .logout()
+          .invalidateHttpSession(true)
+          .deleteCookies("JSESSIONID")
+          .logoutSuccessUrl("http://localhost:3000/")
```
stanislavgrujic: Could you please create a story to extract this value as a parameter, since it will not work on production environment?

PR #32, opened 2018-07-13
B feature/add public holiday service

dujfalusi:


src/main/java/eu/execom/hawaii/service/PublicHolidayService.java
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Retrieves a publicHoliday with a specific id.
+   *
+   * @param id the PublicHoliday id.
+   * @return the PublicHoliday.
+   * @throws EntityNotFoundException if a publicHoliday with given id is not found.
+   */
+  public PublicHoliday getById(Long id) {
+    return publicHolidayRepository.findById(id).orElseThrow(EntityNotFoundException::new);
+  }
```
luinnarn: No need to use `findById(id)`, you can use `getOne(id)`, it throws `EntityNotFoundException`.

src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -24,17 +26,18 @@ public TeamService(TeamRepository teamRepository) {
...
-   * @param id Team id
-   * @return Team with provided id if exists
-   * @throws EntityNotFoundException if a team with given id is not found
+   *
+   * @param id Team id.
+   * @return Team with provided id if exists.
+   * @throws EntityNotFoundException if a team with given id is not found.
    */
   public Team getById(Long id) {
     return teamRepository.findById(id).orElseThrow(EntityNotFoundException::new);
```
luinnarn: See comment above.

src/main/java/eu/execom/hawaii/service/LeaveProfileService.java
```diff
@@ -24,7 +26,7 @@ public LeaveProfileService(LeaveProfileRepository leaveProfileRepository) {
   /**
    * Retrieves a list of all leave profiles from repository.
    *
-   * @return a list of all leave profiles
+   * @return a list of all leave profiles.
    */
   public List<LeaveProfile> getAll() {
     return leaveProfileRepository.findAll();
```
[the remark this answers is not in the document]
dujfalusi: The idea at the first was to have all leave profiles for management by the client. But after discussion this is not a good idea, we will separate get all active and get all inactive... So this will be fixed.

src/main/java/eu/execom/hawaii/service/LeaveProfileService.java
```diff
@@ -33,9 +35,9 @@ public LeaveProfileService(LeaveProfileRepository leaveProfileRepository) {
   /**
    * Retrieves a leave profile with a specific id.
    *
-   * @param id Leave profile id
-   * @return Leave profile with provided id if exists
-   * @throws EntityNotFoundException if a leave profile with given id is not found
+   * @param id Leave profile id.
+   * @return Leave profile with provided id if exists.
+   * @throws EntityNotFoundException if a leave profile with given id is not found.
```
sugrinov: Should we document `Exceptions` not handled by us?

sugrinov: Anyway it's probably documented the same on other Services so I guess it can stay as it is.

dujfalusi: Not necessary to stay because already documented on another place where is throws...

src/main/java/eu/execom/hawaii/service/PublicHolidayService.java
```diff
@@ -0,0 +1,68 @@
...
+  public PublicHoliday getById(Long id) {
+    return publicHolidayRepository.getOne(id);
+  }
+
+  /**
+   * Retrieves all publicHolidays.
+   *
+   * @return a list of publicHolidays.
+   */
+  public List<PublicHoliday> getAll() {
```
sugrinov: Could you rename this to `findAll`?

dujfalusi: This is already discussed earlier, but now will be changed... because have more sense to be findAll if use findAll.

PR #34, opened 2018-07-16
B feature/leave type management

luinnarn:
#33


src/main/java/eu/execom/hawaii/dto/AbsenceDto.java
```diff
@@ -0,0 +1,39 @@
...
+  private boolean active;
+  private boolean deducted;
+
+  public AbsenceDto(Absence absence) {
+    this.id = absence.getId();
+    this.name = absence.getName();
+    this.comment = absence.getComment();
+    this.active = absence.isActive();
+
+    if (absence instanceof Leave) {
```
sugrinov: Since you are extending the abstract class `Absence` maybe you could also create an abstract method (something like getAbsenceType) which would return the type of absence? And then you could do something like
`this.absenceType = absence.getAbsenceType()` instead of checking instances.

src/main/java/eu/execom/hawaii/model/absence/Absence.java
```diff
@@ -21,6 +23,10 @@
 
   private static final long serialVersionUID = -3525779850487542671L;
 
+  @Column(name = "absence_type", insertable = false, updatable = false)
+  @Enumerated(EnumType.STRING)
+  private AbsenceType absenceType;
```
sugrinov: I meant an abstract method but I guess this is ok too :laughing: . Reason for an abstract method was because it forces you to implement it in it's child classes and this makes less room for mistakes.

luinnarn: This simply gets the automatically generated values from the discriminator column.

src/main/java/eu/execom/hawaii/dto/AbsenceDto.java
```diff
@@ -0,0 +1,32 @@
...
+
+import eu.execom.hawaii.model.absence.Absence;
+import eu.execom.hawaii.model.absence.AbsenceType;
+import eu.execom.hawaii.model.absence.Leave;
+import lombok.Data;
+
+import static eu.execom.hawaii.model.absence.AbsenceType.LEAVE;
+
+@Data
+public class AbsenceDto {
```
sugrinov: Will this serialize without `NoArgsConstructor`?

dujfalusi:
Could you please add tests for the service layer?

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Autowired
+  public AbsenceService(AbsenceRepository absenceRepository) {
+    this.absenceRepository = absenceRepository;
+  }
+
+  /**
+   * Retrieves a list of all Absence types from repository.
+   *
+   * @return a list of all Absence types
```
dujfalusi: Please add dots everywhere on the end of the sentence.

src/main/java/eu/execom/hawaii/model/absence/Absence.java
luinnarn: From projectlombok.org:

> @RequiredArgsConstructor generates a constructor with 1 parameter for each 
> field that requires special handling. All non-initialized final fields get a parameter, 
> as well as any fields that are marked as @NonNull that aren't initialized where 
> they are declared. 

Since there are no final or `@NonNull` annotated fields, this basically generates a `NoArgsConstructor`.

src/main/java/eu/execom/hawaii/dto/AbsenceDto.java
luinnarn: From projectlombok.org:

> @RequiredArgsConstructor generates a constructor with 1 parameter for each 
> field that requires special handling. All non-initialized final fields get a parameter, 
> as well as any fields that are marked as @NonNull that aren't initialized where 
> they are declared. 

Since there are no final or `@NonNull` annotated fields, this basically generates a `NoArgsConstructor`.

PR #37, opened 2018-07-19
B feature/request leave

dujfalusi:


src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -63,4 +67,10 @@ public void delete(Long id) {
     absenceRepository.deleteById(id);
   }
 
+  private void checkIsAbsenceLeave(Absence absence) {
+    if (absence.getAbsenceType().equals(AbsenceType.LEAVE)) {
```
sugrinov: What if `absenceType` is null? Maybe it cannot be `null` in this case but it is best practice to always check against a literal (`ENUM`, `constants`, etc) rather then the field itself `AbsenceType.LEAVE.equals(absence.getAbsenceType())`

sugrinov: Also what do you think about moving the comparison directly into the setter?

luinnarn: Another option is a method in the Absence class:
`boolean isLeave() { return AbsenceType.LEAVE.equals(this.absenceType) }`

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -63,4 +67,10 @@ public void delete(Long id) {
...
+      absence.setDeducted(true);
+    }
```
luinnarn: Not all absences of type `LEAVE` are deducted...

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -0,0 +1,97 @@
...
+  public Request save(Request request) {
+    checkUserWithApprover(request);
+    return requestRepository.save(request);
+  }
+
+  private void checkUserWithApprover(Request request) {
+    if (request.getUser().getId().equals(request.getApprover().getId())) {
+      throw new ApproverException();
+    }
+  }
```
luinnarn: Can this situation actually ever happen? Isn't an approverId set only after one of the possible approvers for the user's team actually approves the request? A team can have multiple approvers, none of which can belong to the given team, which effectively aliminates such a possibility...

PR #40, opened 2018-07-20
Implemented session management

luinnarn:


src/main/java/eu/execom/hawaii/api/controller/SignInController.java
```diff
@@ -56,7 +63,13 @@ public SignInController(UserService userService) {
     Authentication authentication = new PreAuthenticatedAuthenticationToken(user, null, authorities);
     SecurityContextHolder.getContext().setAuthentication(authentication);
 
-    return new ResponseEntity<>(HttpStatus.OK);
+    String sessionToken = tokenStore.createToken(user);
+
+    HttpHeaders headers = new HttpHeaders();
+    headers.add(AUTHENTICATION_TOKEN_KEY, sessionToken);
+    headers.add("role", user.getUserRole().name());
```
sugrinov: Why are you sending role in a header? Wouldn't it be better (and easier for frontend devs) to just send the user (needed info such as role, name, etc.) back as response body?

luinnarn: This will be changed once the frontend side is implemented.

sugrinov: Then create an issue for this.

PR #43, opened 2018-07-23
B feature/approve request

dujfalusi:


src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -86,4 +89,22 @@ public Request save(Request request) {
...
+
+    return requestRepository.save(request);
+  }
+
+  private void checkIsApproved(Request request) {
+    if (request.getRequestStatus().equals(RequestStatus.APPROVED)) {
+      allowanceService.applyRequest(request);
+    }
+  }
+
```
luinnarn: You can create a `isApproved()` method in the `Request` class : 
`public boolean isApproved() { return RequestStatus.APPROVED.equals(this.requestStatus) }`
so you can simply have `if (request.isApproved) { allowanceService.applyRequest(request) }`
Also, maybe the method should be called something like `handleRequestStatusUpdate`?

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -0,0 +1,90 @@
...
+    var absence = request.getAbsence();
+
+    switch (absence.getAbsenceType()) {
+      case LEAVE:
+        if (absence.isDeducted()) {
+          deductAnnual(allowance, request);
+        }
+        break;
+      case SICKNESS:
+        incrementSickness(allowance, request);
```
luinnarn: Not sure this should be called `increment`, as increment is usually an addition on a fixed scale, rather simply `add`?

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -0,0 +1,90 @@
...
+        break;
+      case BONUS_DAYS:
+        incrementBonus(allowance, request);
```
luinnarn: See comment above.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -0,0 +1,90 @@
...
+  private void incrementBonus(Allowance allowance, Request request) {
+    allowance.setBonus(allowance.getBonus() + calculateDays(request.getDays()));
+    allowanceRepository.save(allowance);
+  }
+
+  private int calculateDays(List<Day> days) {
+    AtomicInteger numberOfHours = new AtomicInteger(0);
+    days.forEach(day -> increment(day, numberOfHours));
+
+    return numberOfHours.get();
```
luinnarn: This can be simplified:
`private calculateDays(request) {
  var days = request.getDays() //better for debugging purposes
  return days.stream().map(this::getHours).mapToInt(Integer::intValue).sum();
}`
using private method `getHours`:
  `if (day.getDuration().equals(Duration.FULL_DAY)) {
      return 8;
    } else {
      return 4;
    }`
Just probably extract the 8 and 4 to constants.

luinnarn: In fact even simpler, `days.stream().mapToInt(this::getHours).sum();`

luinnarn: Also, the method should be called `calculateHours`, not `calculateDays`...

src/main/java/eu/execom/hawaii/service/AllowanceService.java
sugrinov: It would also be good to add Allowance as suffix since it would make more sense to read `addSicknessAllowance` then `addSickness`. Also this should be done for the other two methods as well.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -0,0 +1,90 @@
...
+        incrementSickness(allowance, request);
+        break;
+      case BONUS_DAYS:
+        incrementBonus(allowance, request);
+        break;
+    }
+  }
+
+  private void deductAnnual(Allowance allowance, Request request) {
+    allowance.setAnnual(allowance.getAnnual() - calculateDays(request.getDays()));
```
sugrinov: Why don't you pass already calculated days instead of passing in request and then taking days out of it.
Something like 
```java
var calculatedDays = calculateDays(requests.getDays());
deductAnnual(allowance, calculatedDays)
```

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -0,0 +1,90 @@
...
+
+  private int calculateDays(List<Day> days) {
+    AtomicInteger numberOfHours = new AtomicInteger(0);
+    days.forEach(day -> increment(day, numberOfHours));
+
+    return numberOfHours.get();
+  }
+
+  private void increment(Day day, AtomicInteger numberOfHours) {
+    if (day.getDuration().equals(Duration.FULL_DAY)) {
```
sugrinov: You could simplify this by removing the else and adding by `4`. And invert the equals to check against the ENUM.
```java
  var daysToAdd = 4;
  if (Duration.FULL_DAY.equals(day.getDuration())) {
    daysToAdd += 4;
  }
 numberOfHours.set(numberOfHours.get() + daysToAdd);
```

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -0,0 +1,90 @@
...
+  Allowance getByUser(User user) {
+    return allowanceRepository.findByUser(user);
+  }
+
+  /**
+   * Apply request after approval on request user allowance.
+   *
+   * @param request the Request.
+   */
+  void applyRequest(Request request) {
```
sugrinov: Why is there no access modifier? It this method package-private?

dujfalusi: yeap, package-private...

src/main/java/eu/execom/hawaii/service/AllowanceService.java
dujfalusi: @luinnarn isn`t ``increment`` more technical term and connected to the number?

dujfalusi: @sugrinov addSicness is already is allowanceService. So is logical to be addSicness to allowance. When you call allowanceService.addSickness()

src/main/java/eu/execom/hawaii/service/RequestService.java
dujfalusi: Yes, this is a good solution when you have multiple check isApproved to avoid repeating same code. From this point of view, I don't see any other place where this could be used. If and when somebody uses again, this check should introduce this method in the model.

sugrinov: At least reverse the equals check to avoid possibility of `NPE`

PR #46, opened 2018-07-25
B feature/request day management

dujfalusi:


src/main/java/eu/execom/hawaii/api/controller/RequestController.java
```diff
@@ -69,27 +77,46 @@ public RequestController(RequestService requestService) {
...
-    request = requestService.save(request);
-    var requestDtoResponse = new RequestDto(request);
+    // Map and save request.
+    var request = mapAndSaveRequest(requestDto);
+    var requestId = request.getId();
+
+    // Map and save request days.
+    requestDto.getDayDtos().forEach(dayDto -> dayDto.setRequestId(requestId));
+    var days = mapDays(requestDto.getDayDtos());
+    days = dayRepository.saveAll(days);
```
sugrinov: Why are we calling repository directly? Shouldn't this be a service call?

sugrinov: Are days going to be saved for every create request? If so this could be called from requestService.save method.

src/main/java/eu/execom/hawaii/api/controller/RequestController.java
```diff
@@ -69,27 +77,46 @@ public RequestController(RequestService requestService) {
...
-    return new ResponseEntity<>(requestDtoResponse, HttpStatus.OK);
+    return new ResponseEntity<>(new RequestDto(request), HttpStatus.OK);
   }
 
   @PutMapping
   public ResponseEntity<RequestDto> handleRequestStatus(@RequestBody RequestDto requestDto) {
+    // Request
     var request = MAPPER.map(requestDto, Request.class);
+    // Request days
+    var days = mapDays(requestDto.getDayDtos());
```
sugrinov: See comment above

PR #50, opened 2018-07-26
B feature/google calendar integration

luinnarn:
Issue #35


src/main/java/eu/execom/hawaii/service/GoogleCalendarService.java
```diff
@@ -0,0 +1,119 @@
...
+    event.setStart(start);
+    event.setEnd(end);
+    event.setSummary("Approved leave: " + day.getRequest().getReason());
+    event.setDescription(day.getRequest().getReason());
+
+    return event;
+  }
+
+  private DateTime getDateTime(LocalDate date, LocalTime localTime) {
+    return new DateTime(Date.from(Instant.from(date.atTime(localTime.atOffset(ZoneOffset.UTC)))),
```
sugrinov: Could you extract this into variables for easier readability?

src/main/java/eu/execom/hawaii/service/GoogleCalendarService.java
```diff
@@ -0,0 +1,119 @@
...
+      case MORNING:
+        start.setDateTime(getDateTime(date, MORNING_TIME));
+        end.setDateTime(getDateTime(date, MID_DAY));
+        break;
+      case AFTERNOON:
+        start.setDateTime(getDateTime(date, MID_DAY));
+        end.setDateTime(getDateTime(date, AFTERNOON_TIME));
+        break;
+      default:
+        start.setDate(new DateTime(true, date.atStartOfDay().toInstant(ZoneOffset.UTC).toEpochMilli(), null));
```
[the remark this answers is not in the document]
sugrinov: Also this should remove the need for the switch. `start.setDate(getDateTime(date, day.getDuration().getStartTime()))` and `end.setDate(getDateTime(date, day.getDuration().getEndTime()))` would be sufficient.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -111,6 +119,11 @@ public Request handleRequestStatusUpdate(Request request) {
   private void checkIsApproved(Request request) {
```
sugrinov: From the name of this method I would think this returns a boolean  after checking if approved or not. My suggestion would be to rename it to something like `applyRequestIfApproved`

sugrinov: I missed this one in the previous PR :speak_no_evil:

src/main/java/eu/execom/hawaii/service/GoogleCalendarService.java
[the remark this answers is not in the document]
luinnarn: The default option creates an all-day event as per requirements. We are using Appogee leave as a reference. An all-day event is created differently, so the time of day is irrelevant, just the date part.

pom.xml
```diff
@@ -70,6 +70,21 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.oauth-client</groupId>
+      <artifactId>google-oauth-client-jetty</artifactId>
+      <version>1.23.0</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api-client</groupId>
+      <artifactId>google-api-client</artifactId>
+      <version>1.23.0</version>
```
dujfalusi: Could you please extract version on top?

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -103,6 +109,8 @@ public Request save(Request mappedRequest) {
    * @return saved request.
    */
   public Request handleRequestStatusUpdate(Request request) {
+    request.setAbsence(absenceService.getById(request.getAbsence().getId()));
```
dujfalusi: If you set absence here, please remove from Request controller. And please create Abstract as a variable for better readability.

src/main/java/eu/execom/hawaii/service/GoogleCalendarService.java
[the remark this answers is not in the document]
sugrinov: Even if time is not relevant this still isn't consistent and by making it consistent you can then simplify it and improve readability.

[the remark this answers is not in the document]
luinnarn: All-day event uses a different method, `setDate` instead of `setDateTime` that uses a different constructor for the `DateTime` object so it can't be made consistent with the previous two.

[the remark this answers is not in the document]
sugrinov: My mistake then, I thought it doesn't matter if the sent date has time or not. Leave it as it is then.

pom.xml
```diff
@@ -70,6 +71,21 @@
       <artifactId>spring-social-google</artifactId>
       <version>1.0.0.RELEASE</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.apis</groupId>
+      <artifactId>google-api-services-calendar</artifactId>
+      <version>v3-rev330-1.23.0</version>
```
sugrinov: Why didn't you extract this one as well?

PR #51, opened 2018-07-26
B feature/view request history by user by date

dujfalusi:


src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -30,6 +34,31 @@ public RequestService(RequestRepository requestRepository, UserRepository userRe
...
   }
 
+  /**
+   * Retrieves a list of request by given dates, ordered by latest.
+   *
+   * @param startDate from date.
+   * @param endDate   to date.
+   * @return a list of requests.
+   */
+  public List<Request> findAllByDates(LocalDate startDate, LocalDate endDate) {
```
luinnarn: See comment above: `findAllWithinDates()`?

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -30,6 +34,31 @@ public RequestService(RequestRepository requestRepository, UserRepository userRe
...
+
+  private Predicate<Request> isBetween(LocalDate startDate, LocalDate endDate) {
+    return request -> checkDates(request, startDate, endDate);
+  }
+
+  private boolean checkDates(Request request, LocalDate startDate, LocalDate endDate) {
+    LocalDate requestDate = request.getDays().get(0).getDate();
+    return (requestDate.isAfter(startDate) || requestDate.isEqual(startDate)) && (requestDate.isBefore(endDate)
+        || requestDate.isEqual(endDate));
+  }
```
luinnarn: If I understood correctly, this will filter only requests that start within `startDate` and `endDate`. What about requests that start before `startDate` but _end_ between `startDate` and `endDate`?

src/main/java/eu/execom/hawaii/api/controller/RequestController.java
```diff
@@ -39,6 +42,16 @@ public RequestController(RequestService requestService, AbsenceService absenceSe
     this.absenceService = absenceService;
   }
 
+  @GetMapping("/dates")
+  public ResponseEntity<List<RequestDto>> getRequestsByDates(
```
luinnarn: Perhaps a better name would be `getRequestsWithinDates`?

PR #57, opened 2018-07-30
Added support for removing events from Google calendar, refactored me…

luinnarn:
Issue #52

### DIFF
--- a/src/main/java/eu/execom/hawaii/service/AllowanceService.java
+++ b/src/main/java/eu/execom/hawaii/service/AllowanceService.java
@@ -40,38 +40,45 @@ public class AllowanceService {
    *
    * @param request the Request.
    */
-  void applyRequest(Request request) {
+  void applyRequest(Request request, boolean apply) {
     var allowance = getByUser(request.getUser());
     var absence = request.getAbsence();
     var days = request.getDays();
+    var hours = calculateHours(days);
+    if (!apply) {
+      hours = -hours;
+    }
 
     switch (absence.getAbsenceType()) {
       case LEAVE:
         if (absence.isDeducted()) {
-          deductAnnual(allowance, days);
+          applyAnnual(allowance, hours);
         }
         break;
       case SICKNESS:
-        incrementSickness(allowance, days);
+        applySickness(allowance, hours);
         break;
       case BONUS_DAYS:
-        incrementBonus(allowance, days);
+        applyBonus(allowance, hours);
         break;
     }
   }
 
-  private void deductAnnual(Allowance allowance, List<Day> days) {
-    allowance.setAnnual(allowance.getAnnual() - calculateHours(days));
+  private void applyAnnual(Allowance allowance, int hours) {
+    var calculatedAnnual = allowance.getAnnual() - hours;
+    allowance.setAnnual(calculatedAnnual);
     allowanceRepository.save(allowance);
   }
 
-  private void incrementSickness(Allowance allowance, List<Day> days) {
-    allowance.setSickness(allowance.getSickness() + calculateHours(days));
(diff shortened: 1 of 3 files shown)


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -40,38 +40,45 @@ Allowance getByUser(User user) {
    *
    * @param request the Request.
    */
-  void applyRequest(Request request) {
+  void applyRequest(Request request, boolean apply) {
```
stanislavgrujic: What does apply mean? What is to be applied? Could you think of a better naming for this parameter?

src/main/java/eu/execom/hawaii/service/GoogleCalendarService.java
```diff
@@ -45,45 +47,78 @@
...
+
+  }
+
+  private Optional<Calendar> getCalendar(Request request) {
+    Calendar calendar = null;
+    try {
+      var httpTransport = GoogleNetHttpTransport.newTrustedTransport();
+      var email = request.getUser().getEmail();
+      var serviceAccountCredentials = new ClassPathResource(CREDENTIALS_FILE_PATH).getFile();
+      var credentialsFromJson = GoogleCredential.fromStream(new FileInputStream(serviceAccountCredentials));
```
stanislavgrujic: This and previous variable can be extracted as a constant, there is no need to read the file each time an email needs to be sent.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
stanislavgrujic: Better naming would be `requestCanceled`.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -151,20 +151,22 @@ public Request handleRequestStatusUpdate(Request request) {
...
-        log.error("Error creating Google calendar event: {}", e);
-      }
-    }
+  private boolean shouldApplyRequest(Request request) {
+    return RequestStatus.APPROVED.equals(request.getRequestStatus()) || RequestStatus.CANCELED.equals(
+        request.getRequestStatus());
+  }
+
+  private void applyRequest(Request request) {
+    boolean apply = RequestStatus.APPROVED.equals(request.getRequestStatus());
```
stanislavgrujic: If you rename this variable to `requestCanceled` and compare the status with CANCELED, you will not need the negation in the `applyRequest` method and the code would be more readable.

pom.xml
```diff
@@ -132,6 +132,11 @@
       <artifactId>modelmapper</artifactId>
       <version>2.1.0</version>
     </dependency>
+      <dependency>
+          <groupId>javax.annotation</groupId>
+          <artifactId>jsr250-api</artifactId>
+          <version>1.0</version>
```
sugrinov: Why don't you use the version provided by spring boot? If you need this specific version please declare it at the top.

src/main/java/eu/execom/hawaii/service/GoogleCalendarService.java
```diff
@@ -44,46 +49,92 @@
...
-
-    request.getDays().stream().map(this::createEvent).forEach(insertEventToCalendar(calendarId, service));
+  void handleRequest(Request request, boolean requestCanceled) {
+
+    Function<Day, Day> setRequest = day -> {
+      day.setRequest(request);
+      return day;
+    };
+
+    Consumer<Calendar> insertEvents = calendar -> request.getDays().stream().map(setRequest)
```
sugrinov: This stream formatting is different from the one below it. Please format them according to your styleguide.

PR #58, opened 2018-07-31
Manage team users and approvers by team update

dujfalusi:


src/main/java/eu/execom/hawaii/repository/TeamApproverRepository.java
```diff
@@ -0,0 +1,12 @@
+package eu.execom.hawaii.repository;
+
+import org.springframework.data.jpa.repository.JpaRepository;
+
+import eu.execom.hawaii.model.Team;
+import eu.execom.hawaii.model.TeamApprover;
+
+public interface TeamApproverRepository extends JpaRepository<TeamApprover, Long> {
```
stanislavgrujic: Why is there a separate entity that acts as a team approver? What was the obstacle to use a regular user instead?

dujfalusi: This separate entity is join table by team and user for team approvers (ManyToMany). Because one team could have more than one approver, and one approver could approve in more than one team.

src/main/java/eu/execom/hawaii/model/User.java
```diff
@@ -13,8 +11,11 @@
 import javax.persistence.OneToMany;
 import javax.validation.constraints.Email;
 import javax.validation.constraints.NotNull;
-import java.io.Serializable;
-import java.util.List;
+
+import eu.execom.hawaii.model.enumerations.UserRole;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
```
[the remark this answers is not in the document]
dujfalusi: We already sync week ago, but this is something before that.

src/main/java/eu/execom/hawaii/repository/TeamApproverRepository.java
stanislavgrujic: Since you don't have any additional information in the join table, you don't need an entity for it. By using ManyToMany annotation with a join table, you can avoid manually having to update all join entity and let the Spring Data / Hibernate do what it is meant to do.
https://vladmihalcea.com/the-best-way-to-use-the-manytomany-annotation-with-jpa-and-hibernate/

dujfalusi: Yes, this simplifies team approvers management a LOT. Thank you for this remark.

src/main/resources/db/changelog/initialDbSchema.xml
```diff
@@ -133,11 +133,12 @@
   </changeSet>
   <changeSet author="admin" id="{initialId}-8">
     <createTable tableName="team_approver">
-      <column autoIncrement="true" name="id" type="BIGINT(19)">
-        <constraints primaryKey="true"/>
+      <column name="team_id" type="BIGINT(19)">
```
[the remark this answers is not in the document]
stanislavgrujic: It is a good remark, but since we are not in production, it would create lots of incremental changes. I suggest to leave it as is, at the moment.

PR #60, opened 2018-08-02
B feature/create user allowance on user creation

dujfalusi:


src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -51,6 +51,7 @@ public UserController(UserService userService) {
   @PostMapping
   public ResponseEntity<UserDto> createUser(@RequestBody UserDto userDto) {
     User user = MAPPER.map(userDto, User.class);
+    userService.createNewUserAllowance(user);
     user = userService.save(user);
```
stanislavgrujic: Why `createUser` does not also save a user in `userService`. It does not make sense to split these two.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -79,4 +86,21 @@ public void delete(Long id) {
...
+  @Transactional
+  public void createNewUserAllowance(User user) {
+    var allowance = new Allowance();
+    var leaveProfile = leaveProfileRepository.getOne(user.getLeaveProfile().getId());
+
+    allowance.setUser(user);
+    allowance.setYear(LocalDate.now().getYear());
+    allowance.setAnnual(leaveProfile.getEntitlement());
+
+    user.setAllowances(new ArrayList<>(Arrays.asList(allowance)));
```
stanislavgrujic: `Arrays.asList()` already creates a new `ArrayList`, no need to do it one more time. Try to know the libraries you are using.

dujfalusi: Arrays.asList() returns immutable list. So I did like this. The list could be managed in future because on this way is mutable.
We currently aren't changing after this ``setAllowance``, so could be immutable for now.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -92,15 +91,17 @@ public void delete(Long id) {
    * @param user new User.
    */
   @Transactional
-  public void createNewUserAllowance(User user) {
+  public User createNewUserAllowance(User user) {
```
stanislavgrujic: Naming is a bit unfortunate, maybe, `createNewAllowanceForUser`or something similar. From current naming, I would expect that `Allowance` is returned from the method, not the altered user.

src/main/java/eu/execom/hawaii/service/UserService.java
sugrinov: You could've used `Lists.newArrayList()` then instead.

PR #62, opened 2018-08-03
B feature/add user years of service

dujfalusi:


src/main/resources/db/changelog/initialDbSchema.xml
```diff
@@ -156,41 +156,41 @@
...
-  <changeSet author="admin" id="{initialId}-18">
+  <changeSet author="admin" id="{initialId}-17">
     <addForeignKeyConstraint baseColumnNames="absence_id" baseTableName="request" constraintName="FKoaypx6rwrdofwxvv0q4e1wn4t" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="absence"/>
   </changeSet>
-  <changeSet author="admin" id="{initialId}-19">
+  <changeSet author="admin" id="{initialId}-18">
     <addForeignKeyConstraint baseColumnNames="approver_id" baseTableName="request" constraintName="FKpbfv7wni1s3f06uoueey4m1he" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="user"/>
   </changeSet>
-  <changeSet author="admin" id="{initialId}-20">
+  <changeSet author="admin" id="{initialId}-19">
```
stanislavgrujic: Why do we have to have a changeset for each of the foreign keys? Is it a good practice? Why can't it be all in a single change set, then you wouldn't have to update it this way.

dujfalusi: This is a recommendation from Liquibase "One Change per ChangeSet".
[Liquibase best practices](http://www.liquibase.org/bestpractices.html)

This file is generated by Liquibase plugin with small changes for author and id.

PR #68, opened 2018-08-07
B feature/add training to allowance

dujfalusi:


src/test/java/eu/execom/hawaii/service/AllowanceServiceTest.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class AllowanceServiceTest {
+
+  @Test
+  public void getByUser() {
+  }
+
+  @Test
+  public void applyRequest() {
+  }
```
stanislavgrujic: I hope this is work in progress... :)

dujfalusi: :) yes

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -53,15 +54,12 @@ void applyRequest(Request request, boolean requestCanceled) {
...
-            applyTraining(allowance, hours);
-            break;
-          default:
-            throw new IllegalArgumentException("Unsupported absence subtype: " + absenceSubtype);
+        if (AbsenceSubtype.ANNUAL.equals(absenceSubtype)) {
+          applyAnnual(allowance, hours);
+        } else if (AbsenceSubtype.TRAINING.equals(absenceSubtype)) {
+          applyTraining(allowance, hours);
+        } else {
+          throw new IllegalArgumentException("Unsupported absence subtype: " + absenceSubtype);
```
stanislavgrujic: Why this change?

src/test/java/eu/execom/hawaii/service/AllowanceServiceTest.java
```diff
@@ -1,14 +1,93 @@
...
+import eu.execom.hawaii.repository.AllowanceRepository;
 
+@RunWith(MockitoJUnitRunner.class)
 public class AllowanceServiceTest {
 
+  @Mock
+  AllowanceRepository allowanceRepository;
+
+  @InjectMocks
+  AllowanceService allowanceService;
```
stanislavgrujic: Please make these fields private.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
dujfalusi: Because subtype is not mandatory. Entering a switch could be null and throw NullPointerExecption

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -49,9 +50,19 @@ void applyRequest(Request request, boolean requestCanceled) {
       hours = -hours;
     }
 
-    switch (absence.getAbsenceType()) {
+    var absenceType = absence.getAbsenceType();
+    var absenceSubtype = absence.getAbsenceSubtype();
+    switch (absenceType) {
       case DEDUCTED_LEAVE:
-        applyAnnual(allowance, hours);
+        if (AbsenceSubtype.ANNUAL.equals(absenceSubtype)) {
```
sugrinov: Can't this also be a switch and have an `Exception` thrown as default case?

dujfalusi: Allways is better to have a proper exception than just Exception... In this case, if the subtype isn't Annual or Training will throw `IllegalArgumentException`.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
stanislavgrujic: And this way, in case of null value for subtype it will throw an IllegalArgumentException, which clashes with the statement that the subtype is not mandatory.

dujfalusi: Yes, isn't mandatory for `Absence` but is mandatory for `Absence` type `DEDUCTED_LEAVE`. I think that is closer to get `IllegalArgumentException` for bad input than 'NullPointerExecption'.

stanislavgrujic: It is not clear from the source code that it is the situation and takes some effort to figure it out. My suggestion is to explicitly check it for null, and if it is so, throw `new IllegalArgumentException("Absence subtype cannot be null.")`, and after that use a switch statement, as was before.

PR #74, opened 2018-08-13
B feature/request approval check approver

dujfalusi:


src/main/java/eu/execom/hawaii/api/controller/RequestController.java
```diff
@@ -87,11 +95,22 @@ public RequestController(RequestService requestService) {
...
-    request = requestService.handleRequestStatusUpdate(request);
+    request = requestService.handleRequestStatusUpdate(request, approver);
 
     return new ResponseEntity<>(new RequestDto(request), HttpStatus.OK);
   }
 
+  private User getUserFromPrincipal(Principal principal) {
+    OAuth2Authentication auth = (OAuth2Authentication) principal;
+    UsernamePasswordAuthenticationToken authUser = (UsernamePasswordAuthenticationToken) auth.getUserAuthentication();
+    LinkedHashMap userDetails = (LinkedHashMap) authUser.getDetails();
```
stanislavgrujic: Please use proper typing here (e.g., <String, Object>).

stanislavgrujic:
Unit tests are missing.

src/main/java/eu/execom/hawaii/api/controller/RequestController.java
dujfalusi: `authUser.getDetails()` returns `Object` if I do this will give me `unchecked cast warning`. Only to `@SuppressWarnings("unchecked")` but this gives a false sense of safety.

src/test/java/eu/execom/hawaii/service/RequestServiceTest.java
```diff
@@ -0,0 +1,310 @@
...
+    absenceAnnual = EntityBuilder.absenceAnnual();
+    absenceTraining = EntityBuilder.absenceTraining();
+
+    dayOne = EntityBuilder.day(LocalDate.of(2018, 11, 20));
+    var dayTwo = EntityBuilder.day(LocalDate.of(2018, 11, 21));
+    var dayThree = EntityBuilder.day(LocalDate.of(2018, 11, 22));
+
+    requestOne = EntityBuilder.request(absenceAnnual, Arrays.asList(dayOne));
+    requestTwo = EntityBuilder.request(absenceTraining, Arrays.asList(dayTwo, dayThree));
+    mockRequests = new ArrayList<>(Arrays.asList(requestOne, requestTwo));
```
stanislavgrujic: Why do you need to wrap a result of `asList()` into a new ArrayList? `asList()` already

src/test/java/eu/execom/hawaii/service/RequestServiceTest.java
```diff
@@ -0,0 +1,310 @@
...
+
+    allMocks = new Object[] {allowanceService, googleCalendarService, requestRepository, userRepository,
```
stanislavgrujic: No need to instantiate a new object before each test. You could do it in the field declaration.

dujfalusi: if I declare a field like this: `Object[] allMocks = new Object[] {allowanceService...}` allMocks will be in this case  `Object[5]` with all elements null. Because @Mock are initialized before each test method.

PR #75, opened 2018-08-13
Email service

dujfalusi:


src/main/resources/application.properties
```diff
@@ -32,3 +33,11 @@ security.oauth2.client.clientAuthenticationScheme=form
 security.oauth2.client.scope=profile email https://www.googleapis.com/auth/calendar
 security.oauth2.resource.userInfoUri=https://www.googleapis.com/userinfo/v2/me
 security.oauth2.resource.preferTokenInfo=false
+
+################### Email configuration ###############################
+spring.mail.host=smtp.gmail.com
+spring.mail.port=587
+spring.mail.username=mpay.noreply@gmail.com
```
stanislavgrujic: Is this correct address?

dujfalusi: Yes, we agreed to use this.

PR #82, opened 2018-08-21
Added get all requests for approval endpoint

dujfalusi:


src/main/java/eu/execom/hawaii/api/controller/RequestController.java
```diff
@@ -45,6 +47,30 @@ public RequestController(RequestService requestService, UserService userService)
...
+    return new ResponseEntity<>(requestDtos, HttpStatus.OK);
+  }
+
+  private List<Request> getRequestsForApprover(List<Team> approverTeams) {
+    return approverTeams.stream()
+                        .map(Team::getUsers)
+                        .flatMap(List::stream)
+                        .map(User::getRequests)
+                        .flatMap(List::stream)
+                        .filter(checkStatus())
```
stanislavgrujic: When you read this line, it would be more expressive if method name describes which requests will you filter out, e.g., `.filter(pendingRequests())`.

PR #84, opened 2018-08-22
B feature/add cancelation pending request status

dujfalusi:


src/main/java/eu/execom/hawaii/api/controller/RequestController.java
```diff
@@ -45,6 +47,30 @@ public RequestController(RequestService requestService, UserService userService)
     this.userService = userService;
   }
 
+  @GetMapping("/approval")
```
dujfalusi: This is code from #82 pull request.

PR #85, opened 2018-08-22
B feature/request cancelation

dujfalusi:


src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -136,52 +136,65 @@ public Request create(Request request) {
...
-  private boolean shouldApplyRequest(Request request) {
-    var isRequestExistingAsApproved = false;
-    if (request.getId() != null) {
-      var existingRequest = requestRepository.getOne(request.getId());
-      isRequestExistingAsApproved = request.getRequestStatus().equals(existingRequest.getRequestStatus());
-    }
-    var isRequestApproved = RequestStatus.APPROVED.equals(request.getRequestStatus());
-    var isRequestCanceled = RequestStatus.CANCELED.equals(request.getRequestStatus());
-    return isRequestApproved || (isRequestCanceled && isRequestExistingAsApproved);
+  private boolean isExistAsApproved(Long requestId) {
```
sugrinov: In my opinion this naming could be simplified to just `isApproved` and also the other one could be `isCancelled`. Since this method doesn't directly check if the request exists but just checks the status of it.

dujfalusi: This isn`t checking from incoming data to be `isApproved` this checks current status from the database with incoming request status. For me, this looks more describable for the given situation.

stanislavgrujic: Stefan has right, since it will not return false if the request does not exist, it will rather throw an exception for that case. See also my comment on the same topic.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -136,52 +136,65 @@ public Request create(Request request) {
...
+            || isCancellationPending(request.getId()))) {
+          applyRequest(request, true);
+        } else if (!isUserRequestApprover(authUser, user) && isApproved(request.getId())) {
+          request.setRequestStatus(RequestStatus.CANCELLATION_PENDING);
+          emailService.createEmailAndSendForApproval(request);
+        }
+        break;
+      default:
+        log.info("Request with status {} will be saved without changing allowance for user with email {}",
+            request.getRequestStatus(), request.getUser().getEmail());
```
stanislavgrujic: It is a good practice to add a `break` statement in the default case as well.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -136,52 +136,65 @@ public Request create(Request request) {
...
-  private boolean shouldApplyRequest(Request request) {
-    var isRequestExistingAsApproved = false;
-    if (request.getId() != null) {
-      var existingRequest = requestRepository.getOne(request.getId());
-      isRequestExistingAsApproved = request.getRequestStatus().equals(existingRequest.getRequestStatus());
-    }
-    var isRequestApproved = RequestStatus.APPROVED.equals(request.getRequestStatus());
-    var isRequestCanceled = RequestStatus.CANCELED.equals(request.getRequestStatus());
-    return isRequestApproved || (isRequestCanceled && isRequestExistingAsApproved);
+  private boolean isApproved(Long requestId) {
```
stanislavgrujic: You should enforce as much type safety as possible. In this case, you could replace `Long` input parameter with `Request` and extract id inside the method. Same should be for `isCancellationPending`.

PR #91, opened 2018-08-29
B feature/manage public holidays weekends on request

dujfalusi:


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -157,6 +165,40 @@ private void applyBonus(Allowance allowance, int hours) {
     allowanceRepository.save(allowance);
   }
 
+  private List<Day> getWorkingDaysOnly(List<Day> days) {
+    var workingDaysWithoutWeekend = getWorkingDaysWithoutWeekend(days);
+
+    LinkedHashSet requestYears = workingDaysWithoutWeekend.stream()
```
stanislavgrujic: Please use `Set` instead. Whenever possible you should use interfaces to define variables, instead of concrete implementations.
Also, add a type to it (from code, I'd say it could be `Set<Integer> requestYears`).

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -157,6 +165,40 @@ private void applyBonus(Allowance allowance, int hours) {
...
+                                                          .map(Day::getDate)
+                                                          .map(LocalDate::getYear)
+                                                          .collect(Collectors.toCollection(LinkedHashSet::new));
+
+    int yearFrom = (int) requestYears.stream().findFirst().get();
+    int yearTo = (int) requestYears.stream().reduce((Object a, Object b) -> b).get();
+    var publicHolidays = publicHolidayRepository.findAllByDateIsBetween(LocalDate.of(yearFrom, 01, 01),
+        LocalDate.of(yearTo, 12, 31));
```
stanislavgrujic: It would be easier to read if you extract these two parameters to variables ->
```
var startOfYearFrom = LocalDate.of(yearFrom, 01, 01);
var endOfYearTo = LocalDate.of(yearTo, 12, 31);
var publicHolidays = publicHolidayRepository.findAllByDateIsBetween(startOfYearFrom, endOfYearTo);
```

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -157,6 +165,40 @@ private void applyBonus(Allowance allowance, int hours) {
...
+
+    var workingDaysOnly = getWorkingDaysWithoutPublicHoliday(workingDaysWithoutWeekend, publicHolidays);
+
+    return workingDaysOnly;
+  }
+
+  private List<Day> getWorkingDaysWithoutPublicHoliday(List<Day> workingDaysWithoutWeekend,
```
stanislavgrujic: Please rename to `getWorkingDatsWithoutPublicHolidays`-> there is certainly more than one holiday.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -157,6 +165,40 @@ private void applyBonus(Allowance allowance, int hours) {
...
+      List<PublicHoliday> publicHolidays) {
+    return workingDaysWithoutWeekend.stream()
+                                    .filter(day -> publicHolidays.stream()
```
stanislavgrujic: Please extract filter predicate to a separate method with a meaningful name (e.g., `Predicate<Day> publicHolidays(List publicHolidays)`).

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -157,6 +165,40 @@ private void applyBonus(Allowance allowance, int hours) {
...
+                                                                 .noneMatch(publicHoliday ->
+                                                                     publicHoliday.getDate().equals(day.getDate())))
+                                    .collect(Collectors.toList());
+  }
+
+  private List<Day> getWorkingDaysWithoutWeekend(List<Day> days) {
+    return days.stream()
+               .filter(day -> !(DayOfWeek.SATURDAY.equals(day.getDate().getDayOfWeek()) || DayOfWeek.SUNDAY.equals(
```
stanislavgrujic: Please extract filter predicate to a separate method with a meaningful name, it will be much easier to read the code.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -168,35 +170,38 @@ private void applyBonus(Allowance allowance, int hours) {
...
 
   private List<Day> getWorkingDaysWithoutWeekend(List<Day> days) {
-    return days.stream()
-               .filter(day -> !(DayOfWeek.SATURDAY.equals(day.getDate().getDayOfWeek()) || DayOfWeek.SUNDAY.equals(
-                   day.getDate().getDayOfWeek())))
-               .collect(Collectors.toList());
+    return days.stream().filter(isNotWeekendDay()).collect(Collectors.toList());
+  }
+
+  private Predicate<Day> isNotWeekendDay() {
```
stanislavgrujic: Please have in mind that for the human brain is easier to think in positive, than negative way, so it is more clear to name this method `isWorkday()` or `workdays()`, and then you can read it in your head *get me a stream of days and filter workdays*, instead  of *get me a stream of days and filter not weekend days*...

stanislavgrujic: You can merge, once you rename the method.

PR #94, opened 2018-09-04
B feature/introduce freemarker template engine

dujfalusi:


src/main/java/eu/execom/hawaii/configuration/SpringAsyncConfig.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.EnableAsync;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+
+@Configuration
+@EnableAsync
+public class SpringAsyncConfig {
+
+  @Bean(name = "threadPoolTaskExecutor")
```
stanislavgrujic: Why do you need to specify the name explicitly? Isn't it the same as the class name by default?

src/main/java/eu/execom/hawaii/util/EmailFormatter.java
```diff
@@ -1,81 +1,20 @@
...
 
-import java.time.LocalDate;
-
 public class EmailFormatter {
 
-  private EmailFormatter() {}
-
-  private static final String FOOTER_MESSAGE = "--------------------------------------------------------\n"
-      + "This is a notification email from the Hawaii HR system. Please do net reply directly to this email.";
+  private EmailFormatter() {
```
stanislavgrujic: This class is not a formatter anymore, it's rather a `EmailSubjectProvider`.

src/main/java/eu/execom/hawaii/util/EmailFormatter.java
```diff
@@ -1,81 +1,20 @@
 package eu.execom.hawaii.util;
 
-import java.time.LocalDate;
-
 public class EmailFormatter {
```
stanislavgrujic: Please make it `final`.

PR #100, opened 2018-09-12
B feature/deduct days from next year allowance

dujfalusi:


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -234,6 +312,21 @@ private int calculateRemainingAnnualHours(Allowance allowance) {
     return totalHours - takenAnnual - pendingAnnual;
   }
 
+  private int calculateRemainingAnnualHoursWithoutPedning(Allowance allowance) {
```
stanislavgrujic: I don't see that pending hours are taken into calculation. Is it a naming issue or calculation issue?

src/test/java/eu/execom/hawaii/service/AllowanceServiceTest.java
```diff
@@ -53,11 +55,11 @@ public void shouldGetAllowanceByUser() {
     given(allowanceRepository.findByUserIdAndYear(mockUser.getId(), 2018)).willReturn(mockAllowance);
 
     // when
-    Allowance allowance = allowanceService.getByUser(mockUser, 2018);
+    Allowance allowance = allowanceService.getByUserByYear(mockUser.getId(), 2018);
```
stanislavgrujic: Please add unit tests for the logic you've added, since it does not seem trivial.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
dujfalusi: Upper method `calculateRemainingAnnualHours` calculate hours with 
`var pendingAnnual = allowance.getPendingAnnual();
return totalHours - takenAnnual - pendingAnnual;`. 

This method calculates without pendingAnnual.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -45,27 +45,28 @@ public AllowanceService(AllowanceRepository allowanceRepository, PublicHolidayRe
...
   /**
    * Apply pending request on request user allowance.
    *
-   * @param request         the Request.
-   * @param pendingCanceled indicate should be pending removed.
+   * @param request the Request.
    */
   @Transactional
-  public void applyPendingRequest(Request request, boolean pendingCanceled) {
+  public void applyPendingRequest(Request request, boolean requestCanceled) {
```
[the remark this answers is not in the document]
dujfalusi: moved

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -136,22 +180,52 @@ public void applyRequest(Request request, boolean requestCanceled) {
...
+    var nextYearRequestedHours = requestedHours - remainingAnnualHoursCurrentYear + nextYearAnnual;
+
+    if (nextYearRequestedHours > 0) {
+      currentYearAllowance.setTakenAnnual(currentYearAnnual + remainingAnnualHoursCurrentYear);
+      nextYearAllowance.setTakenAnnual(nextYearRequestedHours);
+
+      allowanceRepository.save(nextYearAllowance);
+    } else {
+      currentYearAllowance.setTakenAnnual(currentYearAnnual + requestedHours);
+    }
```
stanislavgrujic: You have similar, but not the same, block of code repeated 4 times in this class. I would recommend to make it look exactly the same (by ordering lines inside the if branch the same way), since it will make it easier to read and understand it. Block of code starts at lines 96, 114, 201 and 219. Please make it the same, or similar as much as possible.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -217,8 +214,8 @@ private void applyAnnual(Allowance currentYearAllowance, Allowance nextYearAllow
     var nextYearRequestedHours = requestedHours - remainingAnnualHoursCurrentYear + nextYearAnnual;
 
     if (nextYearRequestedHours > 0) {
-      currentYearAllowance.setTakenAnnual(currentYearAnnual + remainingAnnualHoursCurrentYear);
       nextYearAllowance.setTakenAnnual(nextYearRequestedHours);
+      currentYearAllowance.setTakenAnnual(currentYearAnnual + remainingAnnualHoursCurrentYear);
```
stanislavgrujic: Isn't it more logical to first have current year, and then next year?

PR #103, opened 2018-09-17
B feature/check is request already canceled

dujfalusi:


src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -222,7 +226,10 @@ public Request handleRequestStatusUpdate(Request request, User authUser) {
         applyRequest(request, false);
         break;
       case CANCELED:
-        if (userIsRequestApprover && (requestIsApproved || requestHasPendingCancellation)) {
+        if (requestIsCanceled) {
+          log.error("Request by user: {}, is already canceled.", user.getEmail());
+          throw new EntityExistsException();
```
stanislavgrujic: This is not exactly correct. Please introduce a specific exception, e.g., `RequestAlreadyCanceledException`. Other option is to use `IllegalStateException`. Map it to a proper HTTP status code.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -255,27 +262,22 @@ private boolean isUserRequestApprover(User approver, User requestUser) {
...
-  private boolean isPending(Request request) {
-    var requestId = request.getId();
-    var existingRequest = getById(requestId);
-
+  private boolean isPending(Request existingRequest) {
     return RequestStatus.PENDING.equals(existingRequest.getRequestStatus());
   }
 
+  private boolean isCanceled(Request existingRequest) {
+    return RequestStatus.CANCELED.equals(existingRequest.getRequestStatus());
```
stanislavgrujic: Why not put this internally in `RequestStatus` so that you could use it everywhere as `existingRequest.getRequestStatus().isCanceled(), or you could update Request with this method, so that you have code reading `request.isCanceled()`.

stanislavgrujic: Whatever you decide, apply the same with other statuses of the request.

PR #104, opened 2018-09-17
B fix/provide user information on sign in

dujfalusi:


src/main/java/eu/execom/hawaii/api/controller/SignInController.java
```diff
@@ -45,7 +45,7 @@ public SignInController(UserService userService, TokenStore tokenStore) {
   }
 
   @GetMapping(value = "/signin")
-  public ResponseEntity<Principal> signIn(@RequestHeader(value = "Authorization") String token) {
+  public ResponseEntity<?> signIn(@RequestHeader(value = "Authorization") String token) {
```
stanislavgrujic: Isn't it that you are returning `UserDto`, not an unknown type, as part of ResponseEntity?

dujfalusi: My mistake

PR #118, opened 2018-10-10
B feature/allowance controller get first and last year

OliveraBoskov:
Implemented getting the first and last allowance year for logged in user


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -351,4 +360,23 @@ private void checkRemainingBonusHours(LeaveProfile leaveProfile, Allowance allow
     }
   }
 
+  public Map<String, Integer> getFirstAndLastAllowancesYear(User authUser) {
+//    User user = userRepository.findOne(authUser.getId());
+    List<Allowance> allowances = allowanceRepository.findAllByUserId(authUser.getId());
+    Map<String, Integer> firstAndLastYear = new LinkedHashMap<>();
+    List<Integer> years = new ArrayList<Integer>();
```
stanislavgrujic: Please use diamond operator, instead of explicitly specifying type during typed class instantiation (as is one line above).

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -351,4 +360,23 @@ private void checkRemainingBonusHours(LeaveProfile leaveProfile, Allowance allow
...
+
+    for(Allowance a: allowances){
+      years.add(a.getYear());
+    }
```
stanislavgrujic: Please configure the formatter to properly format your code. This is not properly formatted.

PR #122, opened 2018-11-05
B fix/request service tests

OliveraBoskov:
Removed unnecessary code, fixed tests.

### DIFF
--- a/src/main/java/eu/execom/hawaii/service/SendNotificationsService.java
+++ b/src/main/java/eu/execom/hawaii/service/SendNotificationsService.java
@@ -120,10 +120,11 @@ public class SendNotificationsService {
       conn.setRequestProperty("Authorization", "key=" + authorizationKey);
       conn.connect();
 
-      OutputStream os = conn.getOutputStream();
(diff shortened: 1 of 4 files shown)


src/main/java/eu/execom/hawaii/service/SendNotificationsService.java
```diff
@@ -120,14 +120,14 @@ private void send(String convertedToJson) {
       conn.setRequestProperty("Authorization", "key=" + authorizationKey);
       conn.connect();
 
+      OutputStream os = conn.getOutputStream();
+      OutputStreamWriter osw = new OutputStreamWriter(os, StandardCharsets.UTF_8);
+      osw.write(convertedToJson);
+      osw.flush();
```
stanislavgrujic: Why did you remove try clause? Are you aware of the implications?

OliveraBoskov: By mistake. Previously there was an error because I changed the order of code and when correcting it I forgot to put try clause. I am aware that I need try clause to close the connection.

PR #124, opened 2018-11-14
Team name and team id in userDto.

OliveraBoskov:
In agreement with the mobile team, beside team name we also provide team id in UserDto.

### DIFF
--- a/src/main/java/eu/execom/hawaii/dto/UserDto.java
+++ b/src/main/java/eu/execom/hawaii/dto/UserDto.java
@@ -12,6 +12,7 @@ import java.util.stream.Collectors;
 public class UserDto {
 
   private Long id;
+  private Long teamId;
   private String teamName;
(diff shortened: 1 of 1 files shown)


src/main/java/eu/execom/hawaii/dto/UserDto.java
```diff
@@ -24,6 +25,7 @@
 
   public UserDto(User user) {
     this.id = user.getId();
+    this.teamId = (user.getTeam().getId());
```
stanislavgrujic: When do we use this?

OliveraBoskov: Team name is used when displaying informations about user, to show the name of the team to which user belongs, and team id is used for endpoint, when we are getting all requests by certain team by month (@GetMapping("/team/{teamId}/month")).

PR #129, opened 2018-11-28
Changed activate and deleted fields in User, Team and PublicHoliday

OliveraBoskov:
User: needs to have both activated and deleted field. When user is created, he is initially inactive and is unable to sign in. Also needs to have soft delete.
Team: just needs soft delete, no need for active/inactive.
Public holiday: also just soft delete.

Update: User now has UserStatusType which is enum and can be ACTIVE, INACTIVE or DELETED.
Once the user is created, he is initially inactive. Only active users can sign in. Years of service increment only for active users. Once user stops being an employee, his status turns into deleted.
Only inactive users can become active.
Getting users is now changed to be by UserStatusType, which is not required.


src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -110,9 +110,16 @@ public UserController(UserService userService) {
...
+  public ResponseEntity activateUser(@PathVariable Long id) {
+    userService.activate(id);
+
+    return new ResponseEntity(HttpStatus.OK);
+  }
+
   @DeleteMapping("/{id}")
   public ResponseEntity deleteUser(@PathVariable Long id) {
-    userService.delete(id);
+    userService.deleteUser(id);
```
stanislavgrujic: Why did you rename this? It is consistent with other methods if it does not have suffix `user`.

OliveraBoskov: It is already used only for deleting users, and is refered only to users, inside the method so I thought maybe it would be more explicit.

stanislavgrujic: Rather rename input variable to `userId` and revert back method name to `delete`.

src/main/java/eu/execom/hawaii/dto/UserDto.java
```diff
@@ -20,6 +20,7 @@
   private UserRole userRole;
   private String jobTitle;
   private boolean active;
+  private boolean deleted;
```
stanislavgrujic: Do we have to use both `active` and `deleted` fields? This introduces additional complexity, which I am not sure is needed. E.g., can you activate a deleted user (at the moment this check does not exist). Also, `IdTokenVerifierFilter` does not check if a retrieved user is deleted, only if it is active. When you delete a user you do not deactivate it. As you can see, there is plenty of scenarios where you have to take care of both these flags.

OliveraBoskov: The idea was to provide user entity with both active and deleted fields. Active is for user who is, for example, hired but not active still. That was my case when I got to Execom, they added me on Apogee, but I wasn't able to sign in that first day of work. I don't know how important is that and do we need it?
If we need it, I can add those cheks. And if not, I will leave just flag deleted which is for employees who are no longer working for company.

PR #132, opened 2018-12-13
Redirected all api calls to /api. Using id_token instead of access_token

vrackovic:
- Changed `GetTokenRequest` to initially send `id_token` instead of `access_token`
- Currently using `/users/{email}`to load users data. Probably would be changed to `/me` after first `id_token` validation to gather data about logged in user
- Previous `http://localhost:8080` proxy changed to `http://localhost:8080/api` as requested from changes on backend
- `getTokenSaga` passes whole user object received from google instead of only `access_token`

### DIFF
--- a/src/main/frontend/src/store/services/getTokenRequest.js
+++ b/src/main/frontend/src/store/services/getTokenRequest.js
@@ -1,10 +1,10 @@
 import request from 'superagent';
 
-export const tokenRequest = async accessToken => {
+export const tokenRequest = async userObj => {
   try {
     const response = await request
-      .get('/signin')
-      .set('Authorization', accessToken)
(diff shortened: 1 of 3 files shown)


src/main/frontend/src/store/services/getTokenRequest.js
```diff
@@ -1,10 +1,10 @@
 import request from 'superagent';
 
-export const tokenRequest = async accessToken => {
+export const tokenRequest = async userObj => {
   try {
     const response = await request
-      .get('/signin')
-      .set('Authorization', accessToken)
+      .get(`/users/${userObj.profileObj.email}`)
```
iscekic: Please leave a comment to the BE team to make this better. 💃

PR #138, opened 2018-12-21
B fix/user add start working day

mariobasic:
StartedWorkingDate, and StartedWorkingAtExecomDate fields added to User


mariobasic:
NOTE: in UserService method addServiceYearsToUser will be changed in next pull request (witch requires adding fields to LeaveProfile so that's why I'll do it on different branch)  because LeaveProfile upgrade cannot be based on LeaveProfile Id increment.

stanislavgrujic:

PR #142, opened 2019-01-08
B-fix/pom_fix

JonathanTolic:
plantuml version fix


stanislavgrujic:
Please try to use more expressive commit messages (what was changed in the file) in the future.

PR #143, opened 2019-01-09
B feature/me api

JonathanTolic:
Created the /me api to get current user data. Accessible to all
Fixed Broken tests / reasons for being broke (IdTokenVerifier null checks)
Updated ReadMe to include redux devtools

### DIFF
--- a/src/main/java/eu/execom/hawaii/model/User.java
+++ b/src/main/java/eu/execom/hawaii/model/User.java
@@ -6,13 +6,22 @@ import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
 
-import javax.persistence.*;
+import javax.persistence.CascadeType;
+import javax.persistence.Column;
+import javax.persistence.Lob;
+import javax.persistence.Basic;
+import javax.persistence.FetchType;
+import javax.persistence.Entity;
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+import javax.persistence.ManyToMany;
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToMany;
 import javax.validation.constraints.Email;
 import javax.validation.constraints.NotNull;
 import java.io.Serializable;
 import java.time.LocalDate;
 import java.util.List;
-import java.sql.Blob;
 
 @Entity
 @Data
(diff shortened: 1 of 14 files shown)


stanislavgrujic:
Could you please make the changes we agreed on this morning and push the commit?

JonathanTolic:
already implemented will push along additional changes shortly

src/main/java/eu/execom/hawaii/model/User.java
```diff
@@ -6,20 +6,13 @@
...
-import javax.persistence.CascadeType;
-import javax.persistence.Column;
-import javax.persistence.Entity;
-import javax.persistence.EnumType;
-import javax.persistence.Enumerated;
-import javax.persistence.FetchType;
-import javax.persistence.ManyToMany;
-import javax.persistence.ManyToOne;
-import javax.persistence.OneToMany;
+import javax.persistence.*;
```
stanislavgrujic: Please do not group imports. Configure your IDE to not group it unless there are 50 of it from the same package.

src/main/java/eu/execom/hawaii/model/User.java
```diff
@@ -62,6 +55,11 @@
   @NotNull
   private int yearsOfService;
 
+  @Lob
+  @Basic(fetch = FetchType.LAZY)
```
stanislavgrujic: Isn't FetchType.LAZY by default? Do you have to explicilty set it?

src/main/java/eu/execom/hawaii/model/User.java
```diff
@@ -62,6 +55,11 @@
...
+  @Column(name = "IMAGE", columnDefinition="MEDIUMBLOB")
```
stanislavgrujic: Column name should be by convention set to this name, so no need to overwrite it manually, just use the `columnDefinition` property, and small caps, if possible.

src/main/java/eu/execom/hawaii/model/User.java
JonathanTolic: Afaik, all attributes are fetched eagerly by default, and lazy is used in OneToMany etc. An Since I we only have one piece of code where the blob is used I wanted to reduce how often it is fetched.

src/main/java/eu/execom/hawaii/model/User.java
JonathanTolic: will do, thnx

src/main/java/eu/execom/hawaii/model/User.java
JonathanTolic: will do, thnx

PR #145, opened 2019-01-15
Execom calendar

vrackovic:
- Rendered calendar by using `moment-js` to know current day
- Calendar changes dependent of selected year, or current year if none is selected
- By using metadata, it is determined which days are weekend, today, public holidays, or disabled
- Day selection enabled (currently only cell borders are colored to 'red', intended for later absence requests)


vrackovic:
Included in #148

PR #150, opened 2019-01-22
teamUsers now takes an optional request param, teamId. This is meant …

JonathanTolic:
teamUsers now has an optional request param teamId. Used for the team calendar if the current user is an approver for teams other than the one he is assigned to

Changes to be committed:
	modified:   src/main/java/eu/execom/hawaii/api/controller/UserController.java

### DIFF
--- a/src/main/java/eu/execom/hawaii/api/controller/UserController.java
+++ b/src/main/java/eu/execom/hawaii/api/controller/UserController.java
@@ -8,7 +8,9 @@ import eu.execom.hawaii.model.User;
 import eu.execom.hawaii.model.enumerations.UserStatusType;
 import eu.execom.hawaii.model.Team;
 import eu.execom.hawaii.service.DayService;
+import eu.execom.hawaii.service.TeamService;
 import eu.execom.hawaii.service.UserService;
+import lombok.extern.slf4j.Slf4j;
 import org.modelmapper.ModelMapper;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.domain.Page;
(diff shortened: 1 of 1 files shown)


src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -80,10 +85,21 @@ public UserController(UserService userService, DayService dayService) {
...
       @ApiIgnore @AuthenticationPrincipal User authUser, @RequestParam(required = false) LocalDate startDate,
-      @RequestParam(required = false) LocalDate endDate) {
+      @RequestParam(required = false) LocalDate endDate, @RequestParam(required = false) Long teamId) {
     startDate = assignDefaultStartDate(startDate, endDate);
     endDate = assignDefaultEndDate(startDate, endDate);
-    Team team = authUser.getTeam();
+    Team team;
+    if (teamId != null) {
+      team = teamService.getById(teamId);
+      if (!authUser.getApproverTeams().contains(team) && authUser.getTeam() != team) {
```
stanislavgrujic: Please use `equals()` to compare objects, instead of a reference comparison with `!=`.

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -80,10 +85,21 @@ public UserController(UserService userService, DayService dayService) {
...
+        log.error("User: {} is neither a member of this team nor a team approver for the team {} ", authUser.getEmail(),
```
stanislavgrujic: Not sure if this deserves an error in logs. Maybe put it to debug?
Also, it is a good practice to use `'{}'` (enclose parameter brackets with an apostrophe), so that it is clear what is a parametrized value.

src/main/java/eu/execom/hawaii/api/controller/UserController.java
stanislavgrujic: Could you also resolve this one?

PR #152, opened 2019-01-31
B feature/audit implementation

mariobasic:


src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -164,43 +170,65 @@ private LocalDate assignDefaultEndDate(LocalDate startDate, LocalDate endDate) {
...
+    var oldUser = userService.getUserById(id);
+    UserDto userDto = new UserDto(oldUser);
+    userDto.setUserStatusType(UserStatusType.DELETED);
+    sendAuditInformationForUpdate(OperationPerformed.DELETE, authUser, oldUser, new UserDto(oldUser), userDto);
     userService.delete(id);
+
     return new ResponseEntity(HttpStatus.NO_CONTENT);
   }
 
+  private void sendAuditInformationForUpdate(OperationPerformed operationPerformed, User authUser, User modifiedUser,
```
stanislavgrujic: Move this logic to `UserService`.

src/main/java/eu/execom/hawaii/service/AuditInformationService.java
```diff
@@ -0,0 +1,86 @@
...
+  private String convertToJSON(Object objectForSerialization) {
+    OBJECT_MAPPER.registerModule(new JavaTimeModule());
+    OBJECT_MAPPER.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
+    String currentValue = null;
+
+    try {
+      currentValue = OBJECT_MAPPER.writeValueAsString(objectForSerialization);
+    } catch (JsonProcessingException e) {
+      log.error("Failed to map object of class '{}' as string", objectForSerialization.getClass());
+      e.printStackTrace();
```
stanislavgrujic: Please remove console printing. If you need it, configure the console appender.

src/main/java/eu/execom/hawaii/service/AuditInformationService.java
```diff
@@ -0,0 +1,86 @@
...
+   * Build audit information and save it.
+   *
+   * @param operationPerformed what operation was performed (CREATE, REMOVE, UPDATE, DELETE, ACTIVATE).
+   * @param entity             on witch entity was the operation performed (USER, TEAM, REQUEST).
+   * @param modifiedByUser     user that made change on that entity.
+   * @param modifiedUser       user on witch change was performed.
+   * @param previousState      state of the object before the change.
+   * @param currentState       state of the object after the change.
+   */
+  public void buildAuditInformation(OperationPerformed operationPerformed, AuditedEntity entity, User modifiedByUser,
```
stanislavgrujic: In order to have at least some type safety, you should make `previousState` and `currentState` of type `BaseEntity`, instead of an `Object`.

stanislavgrujic: Also, method name is misleading - I would not expect that it is saved, and would expect that something is returned from a method named `build<Something>`. Maybe rename it to `saveAudit()`.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -303,6 +304,14 @@ public Request handleRequestStatusUpdate(Request request, User authUser) {
     return requestRepository.save(request);
   }
 
+  public void saveAuditInformation(OperationPerformed operationPerformed, User modifiedByUser, Request request,
```
stanislavgrujic: This method should be private and invoked from inside the other method's bodies.

src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -67,10 +82,25 @@ public Team save(Team team) {
    * @param id - the team id.
    */
   @Transactional
-  public void delete(Long id) {
+  public void delete(Long id, User modifiedByUser) {
     var team = getById(id);
+    var previousTeamState = TeamAudit.createTeamAuditEntity(team);
     team.setDeleted(true);
+    saveAuditInformation(OperationPerformed.DELETE, modifiedByUser, team, previousTeamState);
     teamRepository.save(team);
```
stanislavgrujic: Saving audit information should happen only after the changes to the audited entity are successfully saved. Please update wherever applicable in all three services.

PR #157, opened 2019-02-06
Users-GetAll Pagination

danilo996:


src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -58,16 +58,19 @@ public UserController(UserService userService, DayService dayService, TeamServic
   }
 
   @GetMapping
-  public ResponseEntity<List<UserDto>> getUsers(@RequestParam(required = false) List<UserStatusType> userStatusType) {
-    List<User> users;
+  public ResponseEntity<Page<UserDto>> getUsers(@RequestParam(required=false) int page, @RequestParam(required=false) int size,
+      @RequestParam(required=false) List<UserStatusType> userStatusType) {
```
mariobasic: Please apply formatting.

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -58,16 +58,19 @@ public UserController(UserService userService, DayService dayService, TeamServic
...
+    Pageable pageable = PageRequest.of(page, size);
+    Page<User> users;
     if (userStatusType != null && !userStatusType.isEmpty()) {
-      users = userService.findAllByUserStatusType(userStatusType);
+      users = userService.findAllByUserStatusTypePage(userStatusType,pageable);
```
mariobasic: If `page` and `size` are not given, does this return all users?

mariobasic: If `page` and `size` are not given does this return all users?

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -58,16 +58,19 @@ public UserController(UserService userService, DayService dayService, TeamServic
...
     } else {
-      users = userService.findAllUsers();
+      users = userService.findAllUsers(pageable);
```
mariobasic: If `page` and `size` are not given does this return all users?

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -56,12 +56,15 @@ public UserService(UserRepository userRepository, LeaveProfileRepository leavePr
...
    * @return a list of all users, both active and non-active
    */
-  public List<User> findAllUsers() {
-    return userRepository.findAll();
+
+  public Page<User> findAllUsers(Pageable pageable) {
+    return userRepository.findAll(pageable);
   }
 
+
```
mariobasic: Please remove empty line.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -72,6 +75,7 @@ public UserService(UserRepository userRepository, LeaveProfileRepository leavePr
     return userRepository.findAllByUserStatusTypeAndTeam(UserStatusType.ACTIVE, team);
   }
 
+
```
mariobasic: Please remove empty line.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -82,6 +86,16 @@ public UserService(UserRepository userRepository, LeaveProfileRepository leavePr
...
+  /**
+   * Retrieves a list of all users from repository(Pageable)
+   * @param userStatusType what is user status (ACTIVE, INACTIVE or DELETED)
+   * @param pageable the Pageable information about size per page and number of page.
+   * @return list of all users from repository with given status (pageable).
+   */
+  public Page<User> findAllByUserStatusTypePage(List<UserStatusType> userStatusType, Pageable pageable) {
+    return userRepository.findAllByUserStatusType(userStatusType, pageable);
+  }
+
```
mariobasic: Please apply formatting.

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -58,16 +58,26 @@ public UserController(UserService userService, DayService dayService, TeamServic
...
-    List<User> users;
+  public ResponseEntity<Page<UserDto>> getUsers(@RequestParam(required = false) Integer page,
+      @RequestParam(required = false) Integer size,
+      @RequestParam(required = false) List<UserStatusType> userStatusType) {
+    Pageable pageable;
+    if (page != null && size != null) {
+      pageable = PageRequest.of(page, size);
+    } else {
+      pageable = PageRequest.of(0, Integer.MAX_VALUE);
+    }
```
mariobasic: Suggestion, maybe extract this to a method?

stanislavgrujic: You should use the `defaultValue = "0"` as the second parameter for `@RequestParam` annotation, instead of these ugly if/else checks.
Also, for the `size` parameter, do not use `Integer.MAX_VALUE` since it can then load all of the records in the database, which can be, potentially thousands, but rather set the default value to 20 or 30.
After you use the `defaultValues`, your code here will be:
```
Pageable pageable = PageRequest.of(page, size);
```

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -58,16 +58,26 @@ public UserController(UserService userService, DayService dayService, TeamServic
   }
 
   @GetMapping
-  public ResponseEntity<List<UserDto>> getUsers(@RequestParam(required = false) List<UserStatusType> userStatusType) {
-    List<User> users;
+  public ResponseEntity<Page<UserDto>> getUsers(@RequestParam(required = false) Integer page,
+      @RequestParam(required = false) Integer size,
+      @RequestParam(required = false) List<UserStatusType> userStatusType) {
```
stanislavgrujic: It is a good practice to first put other search parameters, and at the end of the method signature put the `page` and `size`. Please update all methods, where you have them defined, accordingly.

src/main/java/eu/execom/hawaii/api/controller/UserController.java
stanislavgrujic: Rather return only ACTIVE users, by default.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -56,10 +56,12 @@ public UserService(UserRepository userRepository, LeaveProfileRepository leavePr
   /**
    * Retrieves a list of all users from repository.
    *
+   * @param pageable the Pageable information about size per page and number of page.
    * @return a list of all users, both active and non-active
    */
-  public List<User> findAllUsers() {
-    return userRepository.findAll();
+
+  public Page<User> findAllUsers(Pageable pageable) {
```
stanislavgrujic: Please rename to `findAll(Pageable pageable)`, since it is implicitly clear that it is users which are being fetched.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -56,10 +56,12 @@ public UserService(UserRepository userRepository, LeaveProfileRepository leavePr
   /**
    * Retrieves a list of all users from repository.
    *
+   * @param pageable the Pageable information about size per page and number of page.
    * @return a list of all users, both active and non-active
    */
-  public List<User> findAllUsers() {
-    return userRepository.findAll();
+
```
stanislavgrujic: Remove this line, otherwise, you have a floating JavaDoc separated from a method.

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -58,16 +58,22 @@ public UserController(UserService userService, DayService dayService, TeamServic
...
-  public ResponseEntity<List<UserDto>> getUsers(@RequestParam(required = false) List<UserStatusType> userStatusType) {
-    List<User> users;
+  public ResponseEntity<Page<UserDto>> getUsers(@RequestParam(required = false, defaultValue="ACTIVE") List<UserStatusType> userStatusType,
+      @RequestParam(required = false, defaultValue = "0") Integer page,
+      @RequestParam(required = false, defaultValue = "30") Integer size
+      ) {
+    Pageable pageable = PageRequest.of(page,size);
+
+    Page<User> users;
     if (userStatusType != null && !userStatusType.isEmpty()) {
```
mariobasic: Does this check make sense with recent changes? Can this` if-else` statement be removed altogether and only left with `users = userService.findAllByUserStatusTypePage(userStatusType, pageable);`? Also please apply formatting.

Issue #161, opened 2019-02-08
Add overflow to the inner div and fix employees table


aberonja:
- Overflow should be added to the inner div of content 
- Employees table should have text overflow fixed

PR #162, opened 2019-02-08
f-enhancement/gh-161-overflow-inner-div

aberonja:
Added inner content overflow auto, fixed employees table text overflow, removed unnecessary styled components

GH issue: #161


src/main/frontend/src/components/tabs/Tabs.js
```diff
@@ -1,22 +1,9 @@
...
-const TabList = styled.div`
-  display: flex;
-  flex-direction: row;
-  width: 100%;
-`;
-
 const Tabs = ({ activeTabIndex, data, handleTabClick }) => (
-  <TabsWrapper>
-    <TabList>
+  <div>
```
iscekic: Extra `div`?

src/main/frontend/src/components/employees/EmployeeInfo.js
```diff
@@ -2,67 +2,46 @@ import React from 'react';
...
-      <EmployeesAdditionalInfo>
-        <p>{jobTitle}</p>
-        <p>{email}</p>
-        <p>Years of Service: {yearsOfService}</p>
-        <Button title="View Profile" />
-      </EmployeesAdditionalInfo>
-    </EmployeeContainer>
+    <div className="align-items-center d-flex justify-content-between mb-3">
+      <EmployeesImage />
+      <TableRow width="15%">
```
iscekic: Would it be more clean to use `first-of-type`?

aberonja: It would, but this is actually second of type, and I'm not sure yet how this will look like, I've just made it like this for presentation purpose.

PR #168, opened 2019-02-12
B fix/team create add approvers

danilo996:


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -88,7 +88,7 @@ public void applyPendingRequest(Request request, boolean requestCanceled) {
    * @param userId Allowance user.
    * @return Allowance.
    */
-  Allowance getByUserAndYear(Long userId, int year) {
+  public Allowance getByUserAndYear(Long userId, int year) {
```
stanislavgrujic: Where do you use this method, which you've made public?

PR #169, opened 2019-02-12
B feature/get user by id

danilo996:


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -88,7 +88,7 @@ public void applyPendingRequest(Request request, boolean requestCanceled) {
    * @param userId Allowance user.
    * @return Allowance.
    */
-  Allowance getByUserAndYear(Long userId, int year) {
+  public Allowance getByUserAndYear(Long userId, int year) {
```
mariobasic: Same comment from previous pull requests apply. Please change to private if you don't use it outside of this class.

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -70,6 +70,13 @@ public UserController(UserService userService, DayService dayService, TeamServic
     return new ResponseEntity<>(userDtos, HttpStatus.OK);
   }
 
+  @GetMapping("/{id}")
+  public ResponseEntity<UserDto> getUserById(@PathVariable Long id) {
+    var user = userService.getUserById(id);
+    var userDto = new UserDto(user);
+    return new ResponseEntity<>(userDto, HttpStatus.OK);
```
mariobasic: Not a mistake but suggestion you could write `return new ResponseEntity<>(new UserDto(user), HttpStatus.OK);` thus eliminating need for line 76. If you pay attention to other methods in this class it's returned this way anywhere where it could be applied.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -196,7 +196,7 @@ public Request save(Request request, User modifiedByUser) {
    * Saves the provided Request to repository.
    * Makes audit of that save.
    *
-   * @param request the Request entity to be persisted.
+   * @param request        the Request entity to be persisted.
```
stanislavgrujic: Please remove extra spaces.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -182,7 +182,7 @@ public Request getById(Long id) {
...
```
stanislavgrujic: Please remove extra spaces.

PR #179, opened 2019-02-18
Remove Team

danilo996:
- Physically deletes team 

- Check if team member list is empty. If there are members in member list, do not allow delete action


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -72,7 +65,7 @@ public Team save(Team team, User modifiedByUser) {
   /**
    * Saves the provided Team to repository.
    *
-   * @param team the Team entity to be persisted.
+   * @param team           the Team entity to be persisted.
```
stanislavgrujic: Please remove these extra spaces. Check all changed files.

stanislavgrujic:

PR #187, opened 2019-02-21
b-feature/merging-frontend-and-backend

aberonja:
Frontend and Backend running on the same port, implemented with frontend-maven-plugin, copies build folder from frontend/build to resources/static folder from which serves frontend


stanislavgrujic:
@iscekic Is it a common practice to put executable node into the source control? I would say it is a bad practice, since it should be part of environment configuration to install node properly.

stanislavgrujic:
Same for yarn.

iscekic:
You're right, usually executables are provided by the environment and if they're not there the build fails (as it should).

We can do some version checks but other than that nothing else.

I scrolled past this stuff down to FE changes so I didn't spot them, my bad. 😄

stanislavgrujic:
@aberonja Once you remove the committed executables (node, yarn...) you can merge the PR.

pom.xml
```diff
@@ -32,7 +32,10 @@
         <liquibase-version>3.6.2</liquibase-version>
         <plantuml-maven-plugin-version>1.2</plantuml-maven-plugin-version>
         <plantuml-version>8059</plantuml-version>
-
+        <frontend-maven-plugin-version>1.6</frontend-maven-plugin-version>
```
iscekic: @stanislavgrujic Are changes in this file ok? I don't speak `maven` that well. 😄

aberonja: @iscekic I was thinking that maybe we should convert it to npm instead of yarn because it's more common and it is already installed alongside node. Also, in our documentation yarn is not a must prerequisite, it is optional.

PR #194, opened 2019-02-25
B fix/update team members

danilo996:


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -73,9 +73,9 @@ public Team save(Team team, User modifiedByUser) {
   @Transactional
   public Team update(Team team, User modifiedByUser) {
     var oldTeam = getById(team.getId());
-    var users = oldTeam.getUsers();
-    users.addAll(team.getUsers());
-    team.setUsers(users.stream().distinct().collect(Collectors.toList()));
+    for(User u: team.getUsers()){
+      u.setTeam(team);
+    }
```
stanislavgrujic: Fix formatting.

PR #200, opened 2019-02-26
Add users upon team creation

mariobasic:


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -76,20 +76,32 @@ public Team getById(Long id) {
...
+      addUsersToNewTeam(team);
+    }
+
     return teamRepository.save(team);
   }
 
+  private void addUsersToNewTeam(Team team) {
+    var users = team.getUsers();
+    team.setUsers(Collections.emptyList());
+    teamRepository.save(team);
```
stanislavgrujic: Why do you need this, when you also invoke the `teamRepository.save(team)` in line 90?

mariobasic: Because we are creating a new team and want to add users to that new team witch does not exist in database yet.
I removed users from that team, persisted new team to database and than updated that team with given users.

Way in witch relationship between team and user entity is modeled 
`@OneToMany(mappedBy = "team", cascade = {CascadeType.PERSIST, CascadeType.MERGE})` 
makes it that if we just write:  
```
public Team create(Team team, User modifiedByUser) {
   saveAuditInformation(OperationPerformed.CREATE, modifiedByUser, team, null);
    team.getUsers().forEach(user -> user.setTeam(team));
    return teamRepository.save(team);
  }
```
exception: `org.hibernate.PersistentObjectException: detached entity passed to persist: eu.execom.hawaii.model.User` is thrown.

Am I missing something? Is there a better or a different way of doing this?

PR #209, opened 2019-03-04
Add message forwarding to client through GlobalExceptionHandler

mariobasic:


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -362,10 +361,17 @@ private void checkRemainingBonusHours(LeaveProfile leaveProfile, Allowance allow
...
 
     if (requestedHours > remainingBonusHours) {
-      throw new InsufficientHoursException();
+      logAndThrowInsufficientHoursException(remainingBonusHours, requestedHours, allowance.getUser().getEmail(), BONUS);
     }
   }
 
+  private void logAndThrowInsufficientHoursException(int remainingHours, int requestedHours, String userEmail,
+      String leaveType) {
+    log.error("Insufficient hours: available {}, requested {}, for user with email {}", remainingHours, requestedHours,
```
stanislavgrujic: Please use ' to separate value from the message, e.g.,
```
log.error("Insufficient hours: available '{}', requested '{}', for user with email '{}'", ...);
```
Update all logging, which you already changed, in this PR.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -362,10 +361,17 @@ private void checkRemainingBonusHours(LeaveProfile leaveProfile, Allowance allow
...
+        userEmail);
+    throw new InsufficientHoursException("You don't have enough " + leaveType + " hours for this request.");
```
stanislavgrujic: You can also internalize message and just pass the needed argument. In this case, it would be:
```
throw new InsufficientHoursException(leaveType);
```
While the exception constructor would be
```
public InsufficientHoursException(String leaveType) {
  super("You don't have enough " + leaveType + " hours for this request.");
}
```

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -7,75 +7,83 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.bind.annotation.ExceptionHandler;
-import org.springframework.web.bind.annotation.ResponseStatus;
-import org.springframework.web.bind.annotation.RestControllerAdvice;
 
 import javax.persistence.EntityExistsException;
 import javax.persistence.EntityNotFoundException;
 import javax.servlet.http.HttpServletRequest;
+import java.nio.file.AccessDeniedException;
```
stanislavgrujic: Did you intentionally change from security exception to a file exception? This does not seem right.

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -7,75 +7,83 @@
...
+    logException(request, ex);
+    return new ResponseEntity<>(Collections.singletonMap("error", ex.getMessage()),
+        HttpStatus.REQUESTED_RANGE_NOT_SATISFIABLE);
   }
 
-  @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR) // 500
-  @ExceptionHandler(Exception.class)
-  public void handleException(HttpServletRequest request, Exception exception) {
-    logException(request, exception);
+  @ExceptionHandler(IllegalArgumentException.class)
```
stanislavgrujic: Why did you introduce this exception handler? How are you sure that `IllegalArgumentException` is always thrown in case of a bad request? Which case do you handle here?

src/main/java/eu/execom/hawaii/service/LeaveProfileService.java
```diff
@@ -47,7 +48,7 @@ public LeaveProfile getById(Long id) {
    */
   public LeaveProfile create(LeaveProfile leaveProfile) {
     if (leaveProfileRepository.existsByLeaveProfileType(leaveProfile.getLeaveProfileType())) {
-      throw new IllegalArgumentException("Profile with same type already exists.");
+      throw new DuplicateEntryException("Profile with same type already exists.");
```
stanislavgrujic: You can use this message inside of the exception, no need to pass it as an argument, at this moment.

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
mariobasic: After consultation with Aleksandar I created separate exception `DuplicateEntryException` so `IllegalArgumentException` is not thrown in that case. Other cases don't have to be sent so I removed that exception handler from `GlobalExceptionHandler`.

src/main/java/eu/execom/hawaii/service/LeaveProfileService.java
stanislavgrujic: Or you can use `EntityExistsException`, which is already handled in the `GlobalExceptionController`.

PR #210, opened 2019-03-05
B feature/maven profiles

mariobasic:


stanislavgrujic:
Did you communicate this change with the PM so that the build system is properly updated?

mariobasic:
Yes I did, but just in case it's not fully understood don't merge PR yet, I will go through changes with him once again to make sure and give you definite response.

mariobasic:
I talked with PM, he made suggestion to make `mvn clean install` build `frontend` by default, and to make specialized profile which does not build `frontend` and left final decision to you. 

Should I make change to build `frontend` with default profile, or leave it the way it is?

stanislavgrujic:
Do as agreed with the PM, since he is also involved in the build process.

pom.xml
```diff
@@ -200,52 +200,77 @@
...
+        <profile>
+            <id>dev</id>
+        </profile>
+        <!-- skip unit test -->
+        <profile>
+            <id>skip-tests</id>
+            <properties>
+                <maven.test.skip>true</maven.test.skip>
+            </properties>
+        </profile>
```
stanislavgrujic: Why do you need this profile? You can always omit tests by passing `-DskipTests` argument to `mvn` call.

PR #212, opened 2019-03-06
f-fix/changed-error-message-source

aberonja:


src/main/frontend/src/store/sagas/HOC/withErrorHandling.js
```diff
@@ -9,11 +10,14 @@ export function withErrorHandling(saga, errorHandler) {
       // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*
       yield* saga(...args);
     } catch (e) {
+      const errorMessage = get(e.response.body, 'error', 'No error message');
```
iscekic: This can still fail. 😄

aberonja: I'm not sure how :D

iscekic: Not every `e` will have a `response.body`.. 👋

Issue #211, opened 2019-03-05
Create absence request


aberonja:
Create modal with requests and add functionality for request creation.

PR #218, opened 2019-03-11
f-enhancement/gh-211-create-absence-request

aberonja:
Created modal for requests and made request functional.

GH issue #211


src/main/frontend/package.json
```diff
@@ -5,11 +5,13 @@
   "dependencies": {
     "bootstrap": "^4.1.1",
     "connected-react-router": "^4.3.0",
+    "date-fns": "^2.0.0-alpha.23",
```
iscekic: When is `v2` coming out?

src/main/frontend/src/components/calendar/CalendarContainer.js
```diff
@@ -35,6 +38,16 @@ export default class CalendarContainer extends Component {
     )
   };
 
+  componentDidMount() {
+    this.props.requestAllowance(this.state.selectedYear);
+  }
+
+  componentDidUpdate(prevProps) {
+    if (prevProps !== this.props) {
+      this.handleYearChange(this.state.selectedYear);
```
iscekic: Should this be in `getDerivedStateFromProps`?

src/main/frontend/src/components/modal/RequestModal.js
```diff
@@ -0,0 +1,196 @@
...
+class RequestModal extends Component {
+  state = {
+    modal: false,
+    isLeave: false,
+    isSickness: false,
+    isBonus: false
+  };
+
+  componentDidUpdate(prevProps) {
+    if (prevProps !== this.props && this.props.modal.shouldClose) {
```
iscekic: I kind of dislike this. Can we achieve the same thing using conditional rendering?

src/main/frontend/src/components/modal/RequestModal.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  toggle = () => {
+    this.setState(prevState => ({
+      modal: !prevState.modal
+    }));
+  };
+
+  resetState = () => {
+    this.setState(() => ({
```
iscekic: No need for callback in `setState`.

src/main/frontend/src/components/modal/RequestModal.js
```diff
@@ -0,0 +1,196 @@
...
+      isLeave: false,
+      isSickness: false,
+      isBonus: false
+    }));
+  };
+
+  requestType = leaveType => {
+    this.setState(() => ({
```
iscekic: No need for callback.

src/main/frontend/src/components/modal/RequestModal.js
```diff
@@ -0,0 +1,196 @@
...
+    return this.props.leaveTypes.filter(
+      leave => leave.absenceType === leaveType
+    );
+  };
+
+  render() {
+    const checkIfSpecificRequest =
+      this.state.isLeave || this.state.isBonus || this.state.isSickness;
+
+    const requestType = this.state.isLeave ? (
```
iscekic: This isn't really readable. Let's do:

```js
getFormProps = () => {
  // get proper leaveTypes and requestAction based on props
  return { leaveTypes, requestAction };
};

// in render
return (
  <RequestForm user={user} allowance={allowance} {...getFormProps()} />
);
```

src/main/frontend/src/components/modal/RequestModal.js
```diff
@@ -0,0 +1,196 @@
...
+  };
+
+  filterLeaveTypes = leaveType => {
+    return this.props.leaveTypes.filter(
+      leave => leave.absenceType === leaveType
+    );
+  };
+
+  render() {
+    const checkIfSpecificRequest =
```
iscekic:
```suggestion
    const isSpecificRequest =
```

src/main/frontend/src/components/modal/requests/RequestForm.js
```diff
@@ -0,0 +1,289 @@
...
+const validationSchema = Yup.object().shape({
+  reason: Yup.string().required(),
+  absence: Yup.object({
+    id: Yup.string().required()
+  })
+});
+
+class RequestForm extends Component {
+  state = {
+    startDate: new Date(),
```
iscekic:
```suggestion
    startDate: moment(),
```

And then we don't need to init `moment` objects all the time. Look into doing this for other methods in this component as well.

src/main/frontend/package.json
aberonja: https://reactdatepicker.com/#example-15

From react-datepicker documentation
```
// Note: Make sure to npm install the right version of date-fns as
// specified in packaged.json. The default one may not be compatiable
// npm install --save date-fns@version
```

Older major release, which is 1.30.1 is not covering this as I can see.

**2.0.0-alpha.23** is the latest release version before beta, and nothing major will change.

src/main/frontend/src/components/modal/requests/RequestForm.js
aberonja: Up until version 1.8.0, this package was using Moment.js. Starting v2.0.0, we switched to using native Date objects to reduce the size of the package.

It throws an error when I try to pass moment date.

src/main/frontend/src/components/calendar/CalendarContainer.js
aberonja: This is being called on component initialization, and `cdu` is being called after something is updated, which I exactly need.

src/main/frontend/src/components/modal/RequestModal.js
aberonja: I'm not sure what you have in mind. This needs to be done in order to close bootstrap modal, that's why I need to set new state.

PR #229, opened 2019-03-12
Check if user exists when creating new user

danilo996:
- check if user with given e-mail already exists in database

- if user exists throw entity exists exception

### DIFF
--- a/src/main/java/eu/execom/hawaii/service/UserService.java
+++ b/src/main/java/eu/execom/hawaii/service/UserService.java
@@ -367,7 +367,7 @@ public class UserService {
   }
 
   private void logAndThrowEntityExistsException(User user) {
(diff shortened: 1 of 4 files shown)


src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -362,4 +366,9 @@ private void saveAuditInformation(OperationPerformed operationPerformed, User mo
     }
   }
 
+  private void logAndThrowEntityExistsException(User user) {
+    log.error("User with email {} already exists.", user.getEmail());
```
stanislavgrujic: Please use single quotes to separate parametrized value from the rest of the text:
```
log.error("User with email '{}' already exists.", user.getEmail());
```

PR #230, opened 2019-03-12
Check if team with given name already exists when creating new team

danilo996:
- if team exists throw entity exists exception


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -144,4 +149,9 @@ private void saveAuditInformation(OperationPerformed operationPerformed, User mo
     }
   }
 
+  private void logAndThrowEntityExistsException(Team team) {
+    log.error("Team with name {} already exists.", team.getName());
```
stanislavgrujic: Please use single quotes to separate parametrized value:
```
log.error("Team with name '{}' already exists.", team.getName());
```

PR #244, opened 2019-03-14
B feature/create allowance when creating new year

danilo996:


src/main/java/eu/execom/hawaii/api/controller/YearController.java
```diff
@@ -36,16 +37,15 @@ public YearController(YearService yearService) {
...
-      yearService.createAllowanceOnCreateYear(newYear);
-      yearService.save(newYear);
+    Year year;
+    if (!yearService.existsByYear(yearDto.getYear())) {
+      year = MAPPER.map(yearDto, Year.class);
+      yearService.save(year);
+      yearService.createAllowanceOnCreateYear(year);
     } else {
-      log.error("Year {} already exists. Will not create new one.", year.getYear());
+      throw new ActionNotAllowedException("This year already exists.");
```
stanislavgrujic: Rather formulate the exception message to *String.format("Year '%n' already exists", yearDto.getYear())*

src/main/java/eu/execom/hawaii/service/YearService.java
```diff
@@ -68,7 +74,12 @@ private Allowance createAllowance(Year createdYear, LeaveProfile leaveProfile, U
     allowance.setYear(createdYear);
     allowance.setAnnual(leaveProfile.getEntitlement());
     allowance.setTraining(leaveProfile.getTraining());
+    allowanceRepository.save(allowance);
 
     return allowance;
   }
+
+  public boolean existsByYear(int year) {
```
stanislavgrujic: Rather rename this method to *yearExists(int year)*, it is much more expressive.

src/main/java/eu/execom/hawaii/service/YearService.java
```diff
@@ -47,18 +48,23 @@ public Year save(Year year) {
...
-      userAllowances.add(allowance);
-      userRepository.save(user);
+    if (createdYear.isActive()) {
+      List<User> activeUsers = userRepository.findAllByUserStatusTypeIn(List.of(UserStatusType.ACTIVE));
+      for (User user : activeUsers) {
+        LeaveProfile leaveProfile = user.getLeaveProfile();
+        var userAllowances = user.getAllowances();
+        Allowance allowance = createAllowance(createdYear, leaveProfile, user);
+        userAllowances.add(allowance);
+        user.setAllowances(userAllowances);
```
stanislavgrujic: You don't have to explicitly set the allowances here, since the reference does not change.

src/main/java/eu/execom/hawaii/api/controller/YearController.java
```diff
@@ -36,16 +37,15 @@ public YearController(YearService yearService) {
...
-    Year newYear = new Year();
-    if (year == null) {
-      newYear = MAPPER.map(yearDto, Year.class);
-      yearService.createAllowanceOnCreateYear(newYear);
-      yearService.save(newYear);
+    Year year;
+    if (!yearService.yearExists(yearDto.getYear())) {
+      year = MAPPER.map(yearDto, Year.class);
+      yearService.save(year);
+      yearService.createAllowanceOnCreateYear(year);
```
stanislavgrujic: In general, you should not have to call two different methods from a single place, from a single class.
This means that if someone calls your method `createAllowanceOnCreateYear`, it will not work without having `save()` before it.

src/main/java/eu/execom/hawaii/api/controller/YearController.java
```diff
@@ -67,6 +67,9 @@ public YearController(YearService yearService) {
   @PutMapping
   public ResponseEntity<YearDto> updateYear(@RequestBody YearDto yearDto) {
     var year = MAPPER.map(yearDto, Year.class);
+    if (year.getAllowances().isEmpty()) {
+      yearService.createAllowanceOnCreateYear(year);
+    }
     year = yearService.save(year);
```
stanislavgrujic: In general, you should not have to call two different methods from a single place, from a single class.
This means that if someone calls your method `save`, it will not work without having `yearService.createAllowanceOnCreateYear()` before it. It should be internilized in a method within `YearService`.

PR #250, opened 2019-03-15
B-feture/gh-249-return-all-years-with-allowance-for-given-user

mariobasic:
Added endpoint in AllowanceController to return all years for which authenticated user has allowance

Gh issue: #249

### DIFF
--- a/src/main/java/eu/execom/hawaii/service/AllowanceService.java
+++ b/src/main/java/eu/execom/hawaii/service/AllowanceService.java
@@ -502,4 +502,10 @@ public class AllowanceService {
 
     return firstAndLastYear;
   }
+
+  public int[] getAllYearsWithAllowanceForUser(User authUser) {
+    List<Allowance> allowances = allowanceRepository.findAllByUserId(authUser.getId());
+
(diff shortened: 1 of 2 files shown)


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -502,4 +502,10 @@ public AllowanceForUserDto getAllowancesForUser(User user) {
 
     return firstAndLastYear;
   }
+
+  public int[] getAllYearsWithAllowanceForUser(User authUser) {
```
stanislavgrujic: Why do you need array here? Cant it be a list?

mariobasic: I am returning int values, and Lists cannot hold primitive values, should I change it to List<Integer> and return it that way?

stanislavgrujic: Yes, Java does autoboxing/unboxing of primitive values, so you can easily switch from one type to another

PR #252, opened 2019-03-15
Absence class change

danilo996:
- removes absenceSubType attribute from the Absence class

- now absences are distinguished on type only(and deducted boolean in case of Leave absence type)

- adds deducted boolean attribute on absence class(describes if leave is deducted or non deducted). It affects only Leave requests(doesn't affect training, sickness and bonus leave requests). If Leave requests absence is not deducted, it means than requested hours won't affect users allowance(it won't be added on pending or taken hours).

- fix absence service and allowance service tests so they are now adapted to the new code flow


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -76,20 +75,21 @@ public void applyPendingRequest(Request request, boolean requestCanceled) {
       hours = -hours;
     }
 
-    var isAbsenceTypeDeductible = AbsenceType.DEDUCTED_LEAVE.equals(absence.getAbsenceType());
-    var isAbsenceTypeAnnual = AbsenceSubtype.ANNUAL.equals(absence.getAbsenceSubtype());
-    var isAbsenceTypeTraining = AbsenceSubtype.TRAINING.equals(absence.getAbsenceSubtype());
+    var isAbsenceTypeLeave = AbsenceType.LEAVE.equals(absence.getAbsenceType());
+    var isAbsenceTypeTraining = AbsenceType.TRAINING.equals(absence.getAbsenceType());
```
stanislavgrujic: Why not remove this variable and just in place of it use `absence.isLeave()` or `absence.isTraining()`?

PR #258, opened 2019-03-18
Absence Service - Check if Absence exists when creating new absence

danilo996:
- fixes existsById null pointer exception

- check if absence with given name already exists when creating new absence

- fixes tests to work with new code flow


src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -79,4 +82,12 @@ public void delete(Long id) {
     absenceRepository.deleteById(id);
   }
 
+  private boolean absenceExists(Absence absence) {
+    if (absence.getId() == null) {
+      return false;
+    } else {
+      return true;
+    }
```
stanislavgrujic: These lines can be replaced with:
```
private boolean absenceExists(Absence absence) {
  return absence.getId() != null;
}
```

What happens in case that absence is null? Is that an option?

Issue #259, opened 2019-03-18
CRUD fixes


aberonja:
CRUD fixes after backend merge

PR #262, opened 2019-03-19
f-fix/gh-259-crud-fixes

aberonja:
- Fixed CRUD errors after back-end changes
- Added confirmation modal for year confirmation

GH issue: #259


src/main/frontend/src/pages/CreateYear.js
```diff
@@ -54,13 +62,20 @@ class CreateYear extends Component {
                 />
               </div>
               <button
-                disabled={isSubmitting}
                 className="btn"
-                onClick={handleSubmit}
+                onClick={
+                  values.active
```
iscekic: Double ternaries are not very readable.

src/main/frontend/src/store/reducers/modalReducer.js
```diff
@@ -1,17 +1,35 @@
...
-import { toggleModal } from '../actions/modalActions';
+import {
+  toggleConfirmationModal,
+  toggleRequestModal
+} from '../actions/modalActions';
 
 export const initialState = {
-  open: false,
-  selectedDay: undefined
+  requestModal: {
```
iscekic: Let's use simpler state for the modals:

```js
{
  type: 'whatever',
  open: true,
  payload: {}
}
```

src/main/frontend/src/store/reducers/modalReducer.js
```diff
@@ -5,31 +5,27 @@ import {
...
 
 const actionHandlers = {
   [toggleRequestModal]: (state, action) => {
     return {
       ...state,
-      requestModal: {
-        selectedDay: action.payload,
-        open: !state.requestModal.open
+      type: 'requestModal',
+      open: !state.open,
```
iscekic: If we ever need to raise a modal from another modal, this will need to be changed. Leave it as it is for now. 💃

PR #275, opened 2019-03-21
Check if team with given name already exists when creating new team

danilo996:
- if exist throw DataIntegrityViolation exception


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -82,14 +83,15 @@ public Team getById(Long id) {
    */
   @Transactional
   public Team create(Team team, User modifiedByUser) {
-
-    if (teamRepository.existsByName(team.getName())) {
-      logAndThrowEntityExistsException(team);
+    try {
+      teamRepository.create(team);
+    } catch (DataIntegrityViolationException e) {
```
stanislavgrujic: I don't see that the database schema is updated so that the unique team name is enforced.

PR #298, opened 2019-03-26
Request controller and service re-factoring

danilo996:
- get requests sorting is now done through service layer using comparator interface(not through repository - we had problems comparing one list of days with another)

- business logic from request controller moved to request service

- request and day service tests fixed to work with new code flow


src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -69,26 +70,50 @@ public RequestService(RequestRepository requestRepository, UserRepository userRe
...
-                   .filter(isBetween(startDate, endDate))
-                   .collect(Collectors.toList());
+  public List<Request> findAllByDateRange(LocalDate date) {
+    var startDate = date.withDayOfMonth(1);
+    var endDate = date.withDayOfMonth(date.lengthOfMonth());
+    List<Request> requests = requestRepository.findAll();
+    return sort(requests).stream().filter(isBetween(startDate, endDate)).collect(Collectors.toList());
+  }
+
+  public List<Request> getRequestsForApprover(List<Team> approverTeams) {
```
stanislavgrujic: Please rename to `getPendingRequestsFromTeams`, it is more accurate.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -194,6 +210,12 @@ public Request getById(Long id) {
     return firstAndLastDate;
   }
 
+  private List<Request> sort(List<Request> requests) {
+    List<Request> modifiableList = new ArrayList<>(requests);
+    modifiableList.sort(Comparator.comparing(request -> request.getDays().get(0).getDate()));
```
stanislavgrujic: Instead of having this method created, create a method
```
private Comparator<Request> byRequestStartingDate() {
  Comparator.comparing(request -> request.getDays().get(0).getDate()));
}
```

Then you can use it in your code as:
```
public List<Request> findAllByAbsenceType(AbsenceType absenceType) {
  List<Request> requests = requestRepository.findAllByAbsenceAbsenceType(absenceType);
	
  requests.sort(byStartingDate());
  return requests;
}
```

danilo996: My first solution was like this one, but some request service tests threw unsupported operation exception. It turned out that i tried to sort an unmodifiable list(requests). My solution was to make another list and copy elements from first list into it and then apply sorting.

PR #305, opened 2019-04-01
f-feature/gh-304-add-year-handle-to-the-request-page

aberonja:
Added year handler to request page.

GH issue: #304


src/main/frontend/src/components/dashboard/YearHandlerContainer.js
```diff
@@ -0,0 +1,190 @@
...
+  };
+
+  componentDidMount() {
+    this.props.requestAllowance(this.state.selectedYear);
+    this.props.setSelectedYear(this.state.selectedYear);
+  }
+
+  componentDidUpdate(prevProps) {
+    //   TODO: This will not work when personalDays are same length in two different years
+    if (prevProps.personalDays.length !== this.props.personalDays.length) {
```
iscekic: Why not just check if the `personalDays` reference is different?

PR #314, opened 2019-04-09
Get user days for queried user

danilo996:
- change user controller 'UserDays' method - if Id is given as parameter in request body and authUser is teamApprover for queriedUser - return all queriedUser leave days for given time period.


src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -113,18 +114,36 @@ public UserController(UserService userService, DayService dayService, TeamServic
...
+    if (userId != null) {
+      authUser = setQueriedUser(authUser, userId);
+    }
     List<Day> days = dayService.getUserAbsencesDays(authUser, startDate, endDate);
     List<DayDto> daysDto = days.stream().map(DayDto::new).collect(Collectors.toList());
 
     return new ResponseEntity<>(daysDto, HttpStatus.OK);
   }
 
+  private User setQueriedUser(User authUser, Long userId) {
```
stanislavgrujic: Method naming is misleading, you are not setting anything. Please pay attention to naming, it helps improve readability, if done correctly. Rather name this method to - `getUser`. Also, first put the parameter which is required for method to get (what are you looking for), and then the `authUser` who asks for it.

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -113,18 +114,38 @@ public UserController(UserService userService, DayService dayService, TeamServic
...
+  private User getUser(Long userId, User authUser) {
+    User user = userService.getUserById(userId);
+    var authUserId = authUser.getId();
+    boolean isApprover = user.getTeam()
+                             .getTeamApprovers()
+                             .stream()
+                             .anyMatch(approver -> approver.getId().equals(authUserId));
+    if (!isApprover) {
+      throw new AccessDeniedException("Only users team approvers can access this information.");
+    } else {
```
stanislavgrujic: No need to have the else branch in this case. The following is equivalent, but a bit more readable. 
```
if (!isApprover) {
  throw new AccessDeniedException("Only users team approvers can access this information.");
} 

return user;
```

PR #320, opened 2019-04-11
Set Icon field to mandatory when creating new leave type

danilo996:


stanislavgrujic:
Please also update the xml scripts for creating the schema.

PR #346, opened 2019-04-12
Set user role to approver when he becomes team approver

danilo996:


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -109,8 +110,18 @@ public Team update(Team team, User modifiedByUser) {
...
     return teamRepository.save(team);
+
+  }
+
+  private void setApproverRoleToUser(Team team) {
+    List<User> approvers = team.getTeamApprovers();
+    approvers.stream().filter(User::hasRoleUser).forEach(user -> {
+      user.setUserRole(UserRole.APPROVER);
+      userService.save(user);
+    });
```
stanislavgrujic: Why do you change user roles on team creation? It is a side effect which I am not sure is desired.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -400,4 +401,13 @@ private void saveAuditInformation(OperationPerformed operationPerformed, User mo
     }
   }
 
+  public void setUserRole(User user) {
```
stanislavgrujic: It is a bit misleading to use the verb *set* since it leads thoughts towards setters and getters. Rethink the naming.
Also, on user update, you will always reset user's role to USER, if he/she is an approver, but does not have any teams assigned. Is that the desired behavior?

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -400,4 +401,13 @@ private void saveAuditInformation(OperationPerformed operationPerformed, User mo
...
+    if (user.hasRoleApprover()) {
+      List<Team> approverTeams = getUserById(user.getId()).getApproverTeams();
+      if (approverTeams.isEmpty()) {
+        user.setUserRole(UserRole.USER);
+      }
+    }
+  }
+
```
stanislavgrujic: Please add unit tests.

PR #349, opened 2019-04-12
Create new request view dto - for team calendar view

danilo996:
- if request type is sickness - name is set to null(sickness name (details) are hidden)


src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -155,10 +155,15 @@ private boolean checkAnyDayWithinDates(List<Day> days, LocalDate startDate, Loca
...
     List<User> teamUsers = team.getUsers();
-    return teamUsers.stream()
-                    .map(user -> findAllByUserWithinDates(startDate, endDate, user.getId()))
-                    .flatMap(List::stream)
-                    .collect(Collectors.toList());
+    List<Request> requests = teamUsers.stream()
+                                      .map(user -> findAllByUserWithinDates(startDate, endDate, user.getId()))
+                                      .flatMap(List::stream)
+                                      .collect(Collectors.toList());
+    requests.stream()
```
stanislavgrujic: It took me some time to figure out what the code bellow does and why it is part of this pull request. Please extract it to a method with a meaningful name.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -159,13 +159,17 @@ private boolean checkAnyDayWithinDates(List<Day> days, LocalDate startDate, Loca
...
-            .forEach(request -> request.getAbsence().setName(null));
+    hideSicknessName(requests);
 
     return requests;
   }
 
+  private void hideSicknessName(List<Request> requests) {
+    requests.stream()
+            .filter(request -> request.getAbsence().getAbsenceType().equals(AbsenceType.SICKNESS))
+            .forEach(request -> request.getAbsence().setName(AbsenceType.SICKNESS.toString()));
```
danilo996: We need absence name for all request types except sickness where we wanted to hide sickness type(name). I'm not sure if this way to do it is a good practice.

PR #351, opened 2019-04-15
Filter out bonus, rejected and canceled requests from individual and team calendar

danilo996:


src/main/java/eu/execom/hawaii/service/DayService.java
```diff
@@ -41,13 +41,13 @@ public DayService(DayRepository dayRepository, RequestRepository requestReposito
...
+                              .filter(withoutCanceledRejectedBonus())
                               .collect(Collectors.toList());
 
     return dayRepository.findAllByRequestInAndDateIsBetween(filteredRequests, startDate, endDate);
   }
 
-  private Predicate<Request> pendingOrApprovedRequests() {
-    return request -> (request.isCancellationPending() || request.isPending() || request.isApproved());
+  private Predicate<Request> withoutCanceledRejectedBonus() {
+    return request -> !(request.isCanceled() || request.isRejected() || request.getAbsence().isBonusDays());
```
stanislavgrujic: You have this condition repeated two times in the code. Could you extract it to a method within the `Request` class and name it appropriately?

PR #363, opened 2019-04-18
Add @NoArgsConstructor annotation to all dto classes

danilo996:


src/main/java/eu/execom/hawaii/dto/DayDto.java
```diff
@@ -29,6 +29,10 @@
 
   private AbsenceType absenceType;
 
+  public DayDto() { // <-  FALI VAM DEFAULT CONSTRUCTOR!!!!! Kako ce spring da napravi ovaj dto ako jedini konstruktor zahteva
```
JonathanTolic: remove this and use NoArgsConstructor to keep things consistent. Also please do not commit comments I left for you ;)

danilo996:
What a hasty mistake. I'm sorry, it won't happen again. :)

PR #365, opened 2019-04-18
Get requests for all users from given team for given month

danilo996:
-if teamId is not given return list of requests for all users for given month

-if month is not given return list of requests for current month


src/main/java/eu/execom/hawaii/api/controller/RequestController.java
```diff
@@ -115,8 +115,8 @@ public RequestController(RequestService requestService, UserService userService)
     return new ResponseEntity<>(requestDtos, HttpStatus.OK);
   }
 
-  @GetMapping("/team/{teamId}/month")
-  public ResponseEntity<List<RequestViewDto>> getTeamRequestsByMonth(@PathVariable Long teamId,
+  @GetMapping("/team/month")
+  public ResponseEntity<List<RequestViewDto>> getTeamRequestsByMonth(@RequestParam(required = false) Long teamId,
```
JonathanTolic: If method params do not fit into a single line, please have a separate line for method name and one for each param followed by an empty line  i.e. :

public ResponseEntity<List<RequestViewDto>> getTeamRequestsByMonth(
  @RequestParam(required = false) Long teamId,

  var requestDtos = requests.stream().map(RequestViewDto::new).collect(Collectors.toList());
  ...

JonathanTolic: p.s. also add appropriate indentation where needed for some reason they are not visible in my comment

PR #367, opened 2019-04-18
f-enhancement/gh-366-filter-days-for-request

aberonja:
GH issue: #366


src/main/frontend/src/components/calendar/calendarUtils.js
```diff
@@ -1,5 +1,5 @@
-import { get } from 'lodash';
-import flow from 'lodash/flow';
+import _flow from 'lodash/flow';
+import _get from 'lodash/get';
```
BraneOpacic21: I like this type of lodash import 😸

Issue #368, opened 2019-04-19
Cancel and update options


aberonja:
- Add possibility to navigate out of edit/create page in administration.

- Save current tab

PR #369, opened 2019-04-19
f-enhancement/gh-368-cancel-and-update-options

aberonja:
Added navigation out of create/update administration menus, saved current tab.

GH issue: #368


src/main/frontend/src/pages/Administration.js
```diff
@@ -1,43 +1,57 @@
...
     );
   }
 }
+
+const mapStateToProps = state => ({
+  activeTabIndex: getActiveTab(state)
+});
+
+const mapDispatchToProps = dispatch =>
+  bindActionCreators({ setActiveTab }, dispatch);
```
BraneOpacic21: Check if bindActionCreators are really needed here:

`const mapDispatchToProps = {
  disableEditItem,
  updateTableData,
  fetchCandidates
};`

.. so you can anyway call that with `this.props.disableEditItem()`

src/main/frontend/src/pages/CreateLeaveProfile.js
```diff
@@ -127,6 +128,9 @@ class CreateLeaveProfile extends Component {
               >
                 Create leave profile
               </button>
+              <NavLink className="w-100" to="/administration">
+                <button className="btn mt-2 w-100">Cancel</button>
+              </NavLink>
```
BraneOpacic21: This should eventually be a one-liner (used in multiple places).

src/main/frontend/src/pages/Administration.js
aberonja: I left a note about it, and it will be changed as soon as I get back from holiday 🐑

src/main/frontend/src/pages/CreateLeaveProfile.js
aberonja: I agree, but all of these files are going to be refactored when I start working on design.

BraneOpacic21: 🥇

src/main/frontend/src/pages/Administration.js
BraneOpacic21: 🐑

Issue #381, opened 2019-04-23
Default sorting when fetching data from database


danilo996:
- users, teams and other entities should be fetched from database sorted by name or first name by default( they are currently sorted by id(this is default behavior))

danilo996:
![](https://github.trello.services/images/mini-trello-icon.png) [Default sorting when fetching data from database](https://trello.com/c/N1q6eZOS/42-default-sorting-when-fetching-data-from-database)

PR #382, opened 2019-04-23
Default sorting when fetching data from database

danilo996:
#381 

- sort users by fullName

- sort teams by team name

- sort public holidays by name

- sort leave types by name

- fix tests to work with new code flow


src/test/java/eu/execom/hawaii/security/ApiSecurityTests.java
```diff
@@ -180,7 +181,7 @@ public void shouldReturnForbiddenStatusCodeForProtectedUrlRequestsWhichRequireHr
   @Test
   public void shouldReturnOkResponseForProtectedUrlRequestsWhichRequireHrManagerAuthorityWhenActiveUserIsHrManager() {
     User user = EntityBuilder.user(EntityBuilder.team());
-    Pageable pageable = PageRequest.of(0, 30);
+    Pageable pageable = PageRequest.of(0, 30, Sort.Direction.ASC, "fullName");
```
JonathanTolic: Make page size a constant and use it instead of numbers

src/test/java/eu/execom/hawaii/service/EntityBuilder.java
```diff
@@ -29,7 +29,7 @@
 
   public static Team team() {
     var team = new Team();
-    team.setName("My team1");
+    team.setName("My team 1");
```
JonathanTolic: use constants instead of strings

PR #393, opened 2019-05-07
Calculate users leave, sick and non deducted days in given month

danilo996:
It was required of us to distinct leave days on deducted and non deducted leaves


src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -145,13 +146,26 @@ private void calculateUsersLeaveDaysInGivenMonth(Page<UserWithDaysDto> pageableU
     pageableUserDtos.stream().forEach(userWithDaysDto -> {
       List<DayDto> filteredDays = userWithDaysDto.getDays()
                                                  .stream()
-                                                 .filter(dayDto -> dayDto.getAbsenceType().equals(AbsenceType.LEAVE)
-                                                     || dayDto.getAbsenceType().equals(AbsenceType.TRAINING))
+                                                 .filter(dayDto -> (dayDto.getAbsenceType().equals(AbsenceType.LEAVE)
```
stanislavgrujic: Extract this to a method with a meaningful name (e.g. *deductedLeave*, *trainingOrLeave*, etc). Does it make sense to check only for deducted leave, without the leave type?

Issue #396, opened 2019-05-07
Get user days for given month


danilo996:
User Id and YearMonth are optional parameters. If user Id is null, return days for logged user. If YearMonth is null, return user days for current month.

PR #398, opened 2019-05-07
B feature/gh 396 get user days for given month

danilo996:
#396


src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -314,6 +321,25 @@ public ResponseEntity deleteUser(@ApiIgnore @AuthenticationPrincipal User authUs
     return new ResponseEntity<>(new UserDto(authUser), HttpStatus.OK);
   }
 
+  private List<DayViewDto> getSicknessDays(UserWithDaysDto userWithDaysDto) {
```
stanislavgrujic: You are using the result of this method only to count the number of elements in it. Why not just return count, instead of a list?

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -314,6 +321,25 @@ public ResponseEntity deleteUser(@ApiIgnore @AuthenticationPrincipal User authUs
...
+    return userWithDaysDto.getDays().stream().filter(dayDto -> dayDto.isDeducted()).collect(Collectors.toList());
+  }
+
+  private List<DayViewDto> getNonDeductedDays(UserWithDaysDto userWithDaysDto) {
+    return userWithDaysDto.getDays()
+                          .stream()
+                          .filter(dayDto -> dayDto.getAbsenceType().equals(AbsenceType.LEAVE) && !dayDto.isDeducted())
+                          .collect(Collectors.toList());
+
+  }
```
stanislavgrujic: Same as these two methods as my comment above.

Issue #407, opened 2019-05-10
Add select picker


aberonja:

aberonja:
![](https://github.trello.services/images/mini-trello-icon.png) [Month picker Team Calendar](https://trello.com/c/DFe0cbKa/52-month-picker-team-calendar)

PR #408, opened 2019-05-10
f-enhancement/gh-407-add-select-picker

aberonja:
There is some code commented out because it will probably be used in the near future.

GH issue: #407


aberonja:
I also know that MonthHandlerContainer and YearHandlerContainer are similar files, that will be merged into one in the near future. 🐑

PR #415, opened 2019-05-10
B-feature/gh-411-import-requests-from-google-calendar

mariobasic:
Import Requests from google calendar

 - Rename Classes used for importing Users to differentiate from Calendar

 - Create job from importing, reading and writing requests from expor

GH issue: #411


src/main/java/eu/execom/hawaii/batch/CalendarProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+  private Duration getDuration(CalendarImportData calendarImportData, LocalDate currentDate) {
+    LocalDateTime startDateTime = calendarImportData.getEventStart();
+    LocalDateTime endDateTime = calendarImportData.getEventEnd();
+    LocalDate startDate = startDateTime.toLocalDate();
+    LocalDate endDate = endDateTime.toLocalDate();
+
+    if (calendarImportData.isAllDayEvent()) {
+      return Duration.FULL_DAY;
+    } else if (startDate.equals(endDate)) {
+      return (startDateTime.getHour() >= NOON) ? Duration.AFTERNOON : Duration.MORNING;
```
stanislavgrujic: Remove the parenthesis, they are useless in this case.

src/main/java/eu/execom/hawaii/batch/CalendarProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+      return (endDateTime.getHour() <= AFTERNOON) ? Duration.MORNING : Duration.FULL_DAY;
+    } else {
+      return Duration.FULL_DAY;
+    }
+  }
+
+  private Request createRequestFromImport(CalendarImportData calendarImportData, List<Day> days) {
+    Optional<User> user = userRepository.findFirstByFullName(getUsersFullName(calendarImportData));
+    if (user.isEmpty()) {
+      return null;
```
stanislavgrujic: It is a bad practice to return null. Could you return an empty request, or something similar?

src/main/java/eu/execom/hawaii/batch/CalendarDbWriter.java
```diff
@@ -0,0 +1,41 @@
...
+  @Autowired
+  public CalendarDbWriter(RequestRepository requestRepository, AllowanceService allowanceService) {
+    this.requestRepository = requestRepository;
+    this.allowanceService = allowanceService;
+  }
+
+  @Override
+  public void write(List<? extends Request> importedRequests) throws Exception {
+    importedRequests.forEach(this::createRequest);
+    importedRequests.forEach(request -> log.debug("Data saved for request '{}'.", request));
```
stanislavgrujic: Why don't you put debug line also in the first loop, end of `createRequest` method?

src/main/java/eu/execom/hawaii/batch/CalendarProcessor.java
```diff
@@ -0,0 +1,133 @@
...
+    CALENDAR_LEAVE_TYPE.put("Approved Leave", RequestStatus.APPROVED);
+    CALENDAR_LEAVE_TYPE.put("Pending Leave", RequestStatus.PENDING);
+    CALENDAR_LEAVE_TYPE.put("Sick", RequestStatus.APPROVED);
+  }
+
+  @Override
+  public Request process(CalendarImportData calendarImportData) throws Exception {
+    List<Day> days = createDaysFromImport(calendarImportData);
+    Request request = createRequestFromImport(calendarImportData, days);
+    log.debug("Converting '{}' into request '{}' and day O '{}' number of days '{}'.", calendarImportData,
```
stanislavgrujic: What does 0 mean in the logs?

src/main/java/eu/execom/hawaii/batch/CalendarProcessor.java
mariobasic: I read it from documentation: 
`To filter a record, one simply returns "null" from the ItemProcessor. The framework will detect that the result is "null" and avoid adding that item to the list of records delivered to the ItemWriter. As usual, an exception thrown from the ItemProcessor will result in a skip.`

Should I find a different way nonetheless?

stanislavgrujic: No, if it is recommended, and needed, by the framework. Please leave it as is.

Issue #417, opened 2019-05-13
Update user - get approver teams and remove team from list if needed


danilo996:
- we need list of team where Employee (User) is team approver

- we should be able to remove team from users approverTeams list when updating user

PR #418, opened 2019-05-13
Get users approver teams and remove from list if needed

danilo996:
#417


src/main/java/eu/execom/hawaii/service/TeamApproverService.java
```diff
@@ -0,0 +1,57 @@
...
+
+    if (!formerApproverTeams.isEmpty()) {
+      formerApproverTeams.forEach(team -> {
+        if (!userIsLastApproverInTeam(team, updatedUser)) {
+          List<User> newTeamApprovers = team.getTeamApprovers()
+                                            .stream()
+                                            .filter(user -> !(user.getId().equals(updatedUser.getId())))
+                                            .collect(Collectors.toList());
+          team.setTeamApprovers(newTeamApprovers);
+          teamRepository.save(team);
```
stanislavgrujic: Extract this code to a method with a meaningful name. It is difficult to figure out what it does this way.

src/main/java/eu/execom/hawaii/service/TeamApproverService.java
```diff
@@ -0,0 +1,57 @@
...
+      });
+    }
+  }
+
+  private boolean userIsLastApproverInTeam(Team team, User approver) {
+    List<User> users = team.getTeamApprovers();
+    return users.stream()
+                .filter(user -> !(user.getId().equals(approver.getId())))
+                .collect(Collectors.toList())
+                .isEmpty();
```
stanislavgrujic: Instead of collecting, you could do the following:
```
return users.stream()
                     .filter(user -> !(user.getId().equals(approver.getId())))
                     .count() == 0;
```
It is more efficient, than to collect to a list, just to check it's size.

src/main/java/eu/execom/hawaii/service/TeamApproverService.java
```diff
@@ -0,0 +1,57 @@
...
+    List<Team> approverTeams = updatedUser.getApproverTeams();
+    formerApproverTeams.removeAll(approverTeams);
+
+    if (!formerApproverTeams.isEmpty()) {
+      formerApproverTeams.forEach(team -> {
+        if (!userIsLastApproverInTeam(team, updatedUser)) {
+          List<User> newTeamApprovers = team.getTeamApprovers()
+                                            .stream()
+                                            .filter(user -> !(user.getId().equals(updatedUser.getId())))
+                                            .collect(Collectors.toList());
```
stanislavgrujic: Why does this have to be so complex? Can't you just remove the user from the list of approvers?
```
team.getTeamApprovers().remove(user);
```

danilo996: I tried doing this way but for some reason user keeps staying in that list.

stanislavgrujic: How do you check equality of User instances?

PR #425, opened 2019-05-13
B-feature/gh-422-get-public-holidays-by-year

danilo996:


src/main/java/eu/execom/hawaii/service/PublicHolidayService.java
```diff
@@ -46,6 +47,22 @@ public PublicHoliday getById(Long id) {
...
+  /**
+   * Retrieves all public holidays in given year sorted by public holiday name.
+   *
+   * @param year year of public holiday.
+   * @return a list of publicHolidays.
+   */
+  public List<PublicHoliday> findAllByYear(Integer year) {
+    LocalDate startDate = LocalDate.of(year, 01, 01);
+    LocalDate endDate = LocalDate.of(year, 12, 31);
+    List<PublicHoliday> publicHolidays = publicHolidayRepository.findAllByDateIsBetween(startDate, endDate);
```
stanislavgrujic: Why do you need it sorted by name? If you really need it, why not sort it right in the database?

PR #430, opened 2019-05-14
Hide sickness name fix - set sickness leave type name to 'Sickness'

danilo996:


src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -177,7 +177,7 @@ private boolean checkAnyDayWithinDates(List<Day> days, LocalDate startDate, Loca
   private void hideSicknessName(List<Request> requests) {
     requests.stream()
             .filter(request -> request.getAbsence().getAbsenceType().equals(AbsenceType.SICKNESS))
-            .forEach(request -> request.getAbsence().setName(AbsenceType.SICKNESS.toString()));
+            .forEach(request -> request.getAbsence().setName("Sickness"));
```
stanislavgrujic: Why hardcode magic string here?

danilo996: Front end requested it. Is this a bad practice? I could format it using String or Character class methods.

Issue #441, opened 2019-05-20
Get all teams for team calendar list


danilo996:
- user is only able to look for team where he is member

- team approver can only get team where he is member and teams where he is approver

- hr manager can see all teams

danilo996:
![](https://github.trello.services/images/mini-trello-icon.png) [Get all teams for team calendar list according to users role and team membership](https://trello.com/c/wgzhUnuG/58-get-all-teams-for-team-calendar-list-according-to-users-role-and-team-membership)

PR #442, opened 2019-05-20
B-feature/gh-441-get-all-teams-for-calendar-list

danilo996:
#441 

- get user teams depending on users role and membership for team calendar list


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -80,6 +80,31 @@ public Team getById(Long id) {
...
+      case USER:
+        teams.add(authUser.getTeam());
+        break;
+      case APPROVER:
+        teams = userService.getUserById(authUser.getId()).getApproverTeams();
+        teams.add(authUser.getTeam());
+        break;
+      case HR_MANAGER:
+        teams = teamRepository.findAll();
+        break;
```
stanislavgrujic: It is a good practice to always add a `default` case, in which you would throw an `IllegalArgumentException`, for unhandled cases which might be added later on.

PR #448, opened 2019-05-20
Calculate total and next year allowance

danilo996:
#447


src/main/java/eu/execom/hawaii/api/controller/AllowanceController.java
```diff
@@ -76,13 +76,15 @@ public AllowanceController(AllowanceService allowanceService, UserService userSe
...
     AllowanceWithoutYearDto allowanceWithoutYearDto = new AllowanceWithoutYearDto(allowance);
-    assignRemainingAllowance(allowance, allowanceWithoutYearDto);
+    calculateAllowance(allowance, allowanceWithoutYearDto);
 
     return new ResponseEntity<>(allowanceWithoutYearDto, HttpStatus.OK);
   }
 
-  private void assignRemainingAllowance(Allowance allowance, AllowanceWithoutYearDto allowanceWithoutYearDto) {
+  private void calculateAllowance(Allowance allowance, AllowanceWithoutYearDto allowanceWithoutYearDto) {
     allowanceWithoutYearDto.setRemainingAnnual(allowanceService.calculateRemainingAnnualHours(allowance));
```
JonathanTolic: Extract this to to a separate mapper class

JonathanTolic: actually a factory would fit this situation better

stanislavgrujic: Please show a sample how you envision it, or add a link to a reference.

JonathanTolic: `AllowanceWithoutYearDto allowanceWithoutYearDto = AllowanceDtoFactory.createAllowance(allowance)`
with this factory calling the service calls in calculateAllowance.

src/main/java/eu/execom/hawaii/restfactory/AllowanceDtoFactory.java
```diff
@@ -0,0 +1,33 @@
...
+  private AllowanceService allowanceService;
+
+  @Autowired
+  public AllowanceDtoFactory(AllowanceService allowanceService) {
+    this.allowanceService = allowanceService;
+  }
+
+  public AllowanceWithoutYearDto createAllowanceWithoutYearDto(Allowance allowance) {
+    AllowanceWithoutYearDto allowanceWithoutYearDto = new AllowanceWithoutYearDto(allowance);
+    calculateAllowance(allowance, allowanceWithoutYearDto);
```
stanislavgrujic: Instead of creating dto here, and passing it to the calculateAllowance method to have side-effects, inline that method here, and remove it. It does not improve readability.

PR #451, opened 2019-05-21
Security Authorization Configuration Fix

danilo996:


src/main/java/eu/execom/hawaii/security/SecurityConfiguration.java
```diff
@@ -46,21 +46,33 @@ protected void configure(HttpSecurity http) throws Exception {
...
-
-              .antMatchers("/api/users/**", "/api/teams/**", "/api/requests/**", "/api/leaveprofiles/**", "/api/import/**", "/api/audit/**").hasAuthority("HR_MANAGER")
+              .antMatchers(getUserEndpoints()).hasAnyAuthority("HR_MANAGER", "USER", "APPROVER")
+              .antMatchers( getApproverEndpoints()).hasAnyAuthority("HR_MANAGER", "APPROVER")
+              .antMatchers(getHrManagerEndpoints()).hasAuthority("HR_MANAGER")
               .antMatchers("/api/**").permitAll()
           .and()
             .addFilterBefore(idTokenVerifierFilter(), AnonymousAuthenticationFilter.class);
     //@formatter:on
   }
```
JonathanTolic: Not what I had in mind, however this way works as well. In this case it would be better not to have endpoints returned from methods but have them as final variables.
What I suggested was something like:
`var x = http.csrf().disable()
        .sessionManagement().sessionCreationPolicy(STATELESS)
          .and()
            .authorizeRequests()
.antMatchers("/api/users/me", "/api/users/image/**").permitAll()
doUserStuff(x);
doApproverStuff(x);
doHrStuff(x)`

.... of course with a lot better naming

Issue #454, opened 2019-05-21
Change logic for processing bonus days


mariobasic:
Logic should be the same as for training days. 
Changes should add pendingBonus, approvedBonus, and bonusManualAdjust. 
When Bonus day is requested pending bonus value increase to requested amount. Upon accepting or rejecting pending bonus request approvedBonus value is requested amount or zero respectively.

PR #456, opened 2019-05-21
B-fix/gh-454-change-bonus-request-calculation-logic

mariobasic:
[#454] Add approvedBonus, pendingBonus and bonusManualAdjust fields to Allowance

 - Change logic for calculating bonus request to mimic training and annual
    requests

 - Add remainingBonus (hours) field in addition to already mentioned
    fields to AllowanceWithoutYearDto

 - When creating new year assign value of bonus hours to bonus field
    based on leave profile to correspond with included changes

 - Add mentioned filed to liquibase schema

 - Fix tests

 - Add default domain to application.properties.example to resolve
    driver name conflict on staging


src/main/java/eu/execom/hawaii/dto/AllowanceWithoutYearDto.java
```diff
@@ -36,6 +40,9 @@ public AllowanceWithoutYearDto(Allowance allowance) {
     this.training = allowance.getTraining();
     this.takenTraining = allowance.getTakenTraining();
     this.pendingTraining = allowance.getPendingTraining();
+    this.approvedBonus = allowance.getApprovedBonus();
+    this.pendingBonus = allowance.getPendingBonus();
+    this.bonusManualAdjust = allowance.getBonusManualAdjust();
```
stanislavgrujic: Please order these in the same order in which they are declared above.

src/main/java/eu/execom/hawaii/model/Allowance.java
```diff
@@ -46,4 +46,10 @@
   private int takenInPreviousYear;
 
   private int pendingInPreviousYear;
+
+  private int approvedBonus;
+
+  private int pendingBonus;
+
+  private int bonusManualAdjust;
```
stanislavgrujic: Adjust fields in `Allowance` and `AllownaceWithoutYearDto` to be in the same order.

PR #458, opened 2019-05-21
B-fix/get-user-request-and-days

danilo996:
- give endpoints access to users according to their roles


danilo996:
#458

danilo996:
#459

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -413,4 +414,18 @@ private void saveAuditInformation(OperationPerformed operationPerformed, User mo
     }
   }
 
+  public User getUserIfLoggedUserHasPermission(Long userId, User authUser) {
```
stanislavgrujic: Could you just name it `getUser`? You don't have to be verbose so much to say that in the name that user has to have a condition. What you can do is create a JavaDoc which says that there is a case where an exception can be thrown. E.g,
```
/***
 * Get user with the given userId.
 * @throws AccessDeniedException if authUser is not approver for the requested userId.
 */
```

PR #472, opened 2019-05-23
f-fix/holiday-weekend-bonus-request

aberonja:


src/main/frontend/src/components/modal/RequestModal.js
```diff
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 import { ModalBody } from 'reactstrap';
 import { bindActionCreators } from 'redux';
 import styled from 'styled-components';
+import ifProp from 'styled-tools/dist/cjs/ifProp';
```
iscekic: Wrong import path.

iscekic:
```suggestion
import { ifProp } from 'styled-tools';
```

PR #511, opened 2019-05-30
Request already handled error messages

danilo996:


src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -439,24 +441,23 @@ public Request handleRequestStatusUpdate(Request request, User authUser) {
...
-        } else if (userIsAuthUser && requestIsApproved) {
+        } else if (userIsLoggedUser && requestIsApproved) {
           request.setRequestStatus(RequestStatus.CANCELLATION_PENDING);
           emailService.createEmailAndSendForApproval(request);
           sendNotificationsService.sendNotificationToApproversAboutSubmittedRequest(request);
         } else if (userIsRequestApprover && (requestIsApproved || requestHasPendingCancellation)) {
           applyRequest(request, true);
           sendNotificationsService.sendNotificationForRequestedLeave(request.getRequestStatus(), user);
-        } else if (requestIsPending && userIsAuthUser) {
+        } else if (requestIsPending && userIsLoggedUser) {
```
stanislavgrujic: Why is the logic not exactly the same here and line 444? Why the boolean values are switched?

Issue #514, opened 2019-05-30
Database text fields character data limit


danilo996:

danilo996:
We should limit every database table field to adequate number of characters and return message if our limits have been exceeded

PR #515, opened 2019-05-30
Database text fields text length limitation

danilo996:
#514 

- set database text fields length

- fix tests


src/main/resources/db/changelog/initialDbSchema.xml
```diff
@@ -7,10 +7,10 @@
             <column autoIncrement="true" name="id" type="BIGINT(19)">
                 <constraints primaryKey="true"/>
             </column>
-            <column name="name" type="VARCHAR(255)">
+            <column name="name" type="VARCHAR(50)">
```
JonathanTolic: do not edit old changeset as this will cause an error. Make a new changeset where you change the lengths

PR #547, opened 2019-06-10
Handle user archive

danilo996:
- set user team to 'Archived Users' when he becomes archived

- clear users approver list when he becomes archived

- add 'editable' boolean field to team entity(true by default for all teams except 'Archived Users' team

- fix tests


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -121,7 +121,7 @@ public Team create(Team team, User modifiedByUser) {
     if (teamRepository.existsByName(team.getName())) {
       throw new EntityExistsException("Team with name " + team.getName() + " already exists.");
     }
-
+    team.setEditable(true);
```
stanislavgrujic: Can't you make a team editable by default, and just for the custom inserted team set it to false?

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -201,15 +205,36 @@ public User save(User user) {
...
+  }
+
+  private void archiveUserIfDeleted(User updatedUser) {
+    if (updatedUser.getUserStatusType().equals(UserStatusType.DELETED)) {
+      archiveUser(updatedUser);
+    }
+  }
+
+  private void validateUserHasTeamAssigned(User updatedUser) {
+    if (!updatedUser.getUserStatusType().equals(UserStatusType.DELETED) && updatedUser.getTeam()
```
stanislavgrujic: Maybe it would be more convenient to add a method to a User class:
```
public boolean isDeleted() {
  return UserStatusType.DELETED.equals(userStatysType);
}
```

Also:
1. Avoid nesting repository calls results as method parameters, but rather extract it to a variable.
2. Why not use equals method on Team, but use additionally getId()?

When all suggestions are applied, you would write the code above:
```
Team archiveTeam = teamRepository.findByName(ARCHIVE_TEAM_NAME);
if (!updatedUser.isDeleted()) && updatedUser.getTeam().equals(archiveTeam)) {
  throw new IllegalArgumentException("To restore a user, you need to set his team first.");
}
```

src/main/resources/data.sql
```diff
@@ -0,0 +1,138 @@
...
+SET foreign_key_checks = 0; ALTER TABLE leave_profile CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci; SET foreign_key_checks = 1;
+SET foreign_key_checks = 0; ALTER TABLE public_holiday CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci; SET foreign_key_checks = 1;
+SET foreign_key_checks = 0; ALTER TABLE request CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci; SET foreign_key_checks = 1;
+SET foreign_key_checks = 0; ALTER TABLE team CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci; SET foreign_key_checks = 1;
+SET foreign_key_checks = 0; ALTER TABLE team_approver CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci; SET foreign_key_checks = 1;
+SET foreign_key_checks = 0; ALTER TABLE user CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci; SET foreign_key_checks = 1;
+SET foreign_key_checks = 0; ALTER TABLE user_push_token CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci; SET foreign_key_checks = 1;
+SET foreign_key_checks = 0; ALTER TABLE year CONVERT TO CHARACTER SET utf8 COLLATE utf8_general_ci; SET foreign_key_checks = 1;
```
stanislavgrujic: Why is this in the data script?

danilo996: We needed to change default collation for our database in order to support Serbian Latin characters. I could create new XML file where i can write a SQL syntax which changes collation for whole database table. Then i could include that file in master change log. That way there would be no need to change every table individually.

stanislavgrujic: This should go into schema creation, not in the data script. Can't you update the initial script to set collation right when the tables are created?

danilo996: Yes, i will do that. Thank you for suggestion. It is much clearer than my solution.

src/main/java/eu/execom/hawaii/service/UserService.java
stanislavgrujic: OK, you can leave that part then. Please apply other parts of my comment to your code.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -201,15 +205,34 @@ public User save(User user) {
...
   }
 
+  private void forbidLoggedUserToChangeHisRole(User updatedUser, User modifiedByUser) {
+    if (updatedUser.equals(modifiedByUser) && !updatedUser.getUserRole().equals(modifiedByUser.getUserRole())) {
+      throw new IllegalArgumentException("You are not allowed to change your own role.");
+    }
+  }
+
+  private void archiveUserIfDeleted(User updatedUser) {
+    if (updatedUser.getUserStatusType().equals(UserStatusType.DELETED)) {
```
stanislavgrujic: Why not use the method you created `if (updatedUser.isDeleted()) {`?

PR #556, opened 2019-06-13
B feature/gh 553 search teams by team name

danilo996:


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -25,6 +28,8 @@
 @Service
 public class TeamService {
 
+  private static final int MINIMUM_QUERY_SIZE = 3;
```
stanislavgrujic: Rather name this `MINIMUM_QUERY_LENGTH`.

PR #593, opened 2019-06-25
B-fix/check-if-user-already-has-allowance-assigned

danilo996:


src/main/java/eu/execom/hawaii/batch/AllowanceProcessor.java
```diff
@@ -58,18 +69,19 @@ private Allowance createAllowanceFromImport(AllowanceImportData allowanceImportD
     return allowance;
   }
 
+  private boolean checkIfUserAlreadyHasAllowanceAssigned(User user, Year year) {
```
stanislavgrujic: Rename this method to `userHasAlreadyAllowanceAssigned` - it is much easier to read the code.

PR #599, opened 2019-06-26
B fix/check if allowance is divisible by four

danilo996:


src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -521,12 +521,34 @@ public AllowanceForUserDto getAllowancesForUser(User user) {
...
   @Transactional
   public Allowance update(AllowanceEditDto updatedAllowance) {
-    Allowance allowance = allowanceRepository.getOne(updatedAllowance.getId());
+    validateAllowanceInputValues(calculateTotalInputAllowance(updatedAllowance));
+    return allowanceRepository.save(updateAllowanceValues(updatedAllowance));
+  }
+
+  void validateAllowanceInputValues(int inputValue) {
+    if (!divisibleByFour(inputValue)) {
+      throw new IllegalArgumentException(
```
stanislavgrujic: You should create a custom exception here, to overcome the issue with to generic handling of `IllegalArgumentException`, in the `GlobalExceptionHandler`. This new exception could extend the `IllegalArgumentException`.

src/main/java/eu/execom/hawaii/exceptions/NotValidAllowanceValueException.java
```diff
@@ -0,0 +1,10 @@
+package eu.execom.hawaii.exceptions;
+
+public class NotValidAllowanceValueException extends IllegalArgumentException {
```
stanislavgrujic: Rather use *InvalidAllowanceValueException*, sounds better. :)

PR #602, opened 2019-06-26
B-feature/gh-597-implement-approval flow (Request Audit)

mariobasic:
[#597] Change information that is audited for request update 

 - Add endpoint for getting approval flow for request with given id


src/main/java/eu/execom/hawaii/service/ApprovalFlowService.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public List<RequestAudit> getApprovalFlowForRequest(String id) {
+    List<AuditInformation> requestsForApprovalFlow = auditInformationRepository.getAllByAuditedEntityAndCurrentValueContainingOrderByModifiedDateTimeDesc(
+        AuditedEntity.REQUEST, REQUEST_ID_PREFIX + id);
+
+    return requestsForApprovalFlow.stream().map(JSONToRequest()).collect(Collectors.toList());
+  }
+
+  private Function<AuditInformation, RequestAudit> JSONToRequest() {
```
stanislavgrujic: Camel case notation recommends to use small caps also in abbreviations, if it is first word. Please rename to `jsonToRequest`.

src/main/java/eu/execom/hawaii/service/RequestService.java
```diff
@@ -522,14 +522,65 @@ private void logAndThrowNotAuthorizedApprovalException(User user, String handleA
...
+      comment = currentReason;
+
+    } else if (!currentApproverComment.equals(previousApproverComment)) {
+      comment = currentApproverComment;
+
+    } else if (!currentCancellationReason.equals(previousCancellationReason)) {
+      comment = currentCancellationReason;
+
+    } else {
+      comment = Optional.ofNullable(currentRequestState.getApproverCancellationComment());
```
stanislavgrujic: Instead of defining Optional<String> comment at the top of the method, you can just return it inside the if block, such as:
```
if (!currentReason.equals(previousReason)) {
  return currentReason;
} else if (!currentApproverComment.equals(previousApproverComment)) {
  return currentApproverComment;
} else if (!currentCancellationReason.equals(previousCancellationReason)) {
  return currentCancellationReason;
} else {
  return Optional.ofNullable(currentRequestState.getApproverCancellationComment());
}
```

PR #610, opened 2019-07-02
B-fix/audit-information

danilo996:
- add 'modified entity id' field to audit information entity

- save update info for users when creating and updating team(if user has changed team)

- fix tests


src/main/java/eu/execom/hawaii/service/ApprovalFlowService.java
```diff
@@ -28,11 +26,9 @@ public ApprovalFlowService(AuditInformationService auditInformationService,
     this.auditInformationRepository = auditInformationRepository;
   }
 
-  public List<RequestAudit> getApprovalFlowForRequest(String id) {
-    List<AuditInformation> requestsForApprovalFlow = auditInformationRepository.getAllByAuditedEntityAndCurrentValueContainingOrderByModifiedDateTime(
-        AuditedEntity.REQUEST, REQUEST_ID_PREFIX + id);
-
-    return requestsForApprovalFlow.stream().map(jsonToRequest()).collect(Collectors.toList());
+  public List<RequestAudit> getApprovalFlowForRequest(Long id) {
+    return auditInformationRepository.getAllByAuditedEntityAndModifiedEntityIdOrderByModifiedDateTime(
```
stanislavgrujic: This is very difficult to read. Please reformat to:
```
return auditInformationRepository
               .getAllByAuditedEntityAndModifiedEntityIdOrderByModifiedDateTime(AuditedEntity.REQUEST, id)
               .stream()
               .map(jsonToRequest())
               .collect(Collectors.toList());
```

src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -202,19 +206,34 @@ private void logAndThrowActionNotAllowedException(Team team) {
...
-          .forEach(modifiedUser -> auditInformationService.saveAudit(operationPerformed, modifiedByUser, modifiedUser,
-              previousTeamState, currentTeamState));
+  private void saveAuditForUpdatedUsers(Team updatedTeam, User modifiedByUser, Team team) {
+    var newMembers = getNewMembers(updatedTeam, team);
+    saveAuditForUpdatedUsers(modifiedByUser, newMembers);
+  }
+
+  private void saveAuditForUpdatedUsers(User modifiedByUser, List<User> users) {
+    if (!users.isEmpty()) {
+      users.forEach(user -> userService.saveAuditInformation(OperationPerformed.UPDATE, modifiedByUser, user,
```
stanislavgrujic: Why not move `UserAudit.fromUser(user)` calculation inside the `saveAuditInformation()` method body? That way you would have one parameter less in method call.

src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -202,19 +206,34 @@ private void logAndThrowActionNotAllowedException(Team team) {
...
   }
 
+  private List<User> getNewMembers(Team updatedTeam, Team team) {
+    List<User> newTeamMembers = new ArrayList<>();
+    updatedTeam.getUsers().forEach(user -> {
+      if (!team.getUsers().contains(user)) {
+        newTeamMembers.add(user);
+      }
+    });
+    return newTeamMembers;
```
stanislavgrujic: Please use streams at its full capacity:
```
return updatedTeam.getUsers.filter(user -> !team.getUsers().contains(user)).collect(Collectors.toList());
```

src/test/java/eu/execom/hawaii/service/TeamServiceTest.java
```diff
@@ -134,4 +135,32 @@ public void shouldDeleteTeam() {
     verifyNoMoreInteractions(allMocks);
   }
 
+  @Test
+  public void shouldUpdateTeam() {
+    // given
+    var members = new ArrayList<>(List.of(EntityBuilder.user(mockTeam), EntityBuilder.leavingUser()));
```
stanislavgrujic: No need to wrap inside the `ArrayList` - `List.of()` already returns an instance of an `ArrayList`.

src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -205,13 +208,25 @@ private void logAndThrowActionNotAllowedException(Team team) {
...
+
+  private List<User> getNewMembers(Team updatedTeam, Team team) {
+    return updatedTeam.getUsers()
+                      .stream()
+                      .filter(member -> !team.getUsers().contains(member))
+                      .collect(Collectors.toList());
+  }
+
+  private void saveAuditForUpdatedUsers(User modifiedByUser, List<User> users) {
+    if (!users.isEmpty()) {
```
stanislavgrujic: Why do you need to check if it is not empty? If it is not empty, it will just not iterate.

src/test/java/eu/execom/hawaii/service/TeamServiceTest.java
danilo996: These tests were my first task when i came to Execom. I will surely re-factor whole class in near future :-)

PR #615, opened 2019-07-03
B-fix/preserve-ordering-for-auditInformation

mariobasic:
Make `auditInformationMap` variable `TreeMap` instance to preserve ordering


src/main/java/eu/execom/hawaii/model/AuditInformation.java
```diff
@@ -35,4 +35,8 @@
   private String previousValue;
 
   private String currentValue;
+
+  public int compareTo(AuditInformation auditInformation) {
```
stanislavgrujic: Always add `@Override` annotation when you override/implement methods from interfaces or parent classes.

mariobasic:
Reformatting will be done in next PR to make this PR less polluted.

PR #635, opened 2019-07-09
B feature/gh 618 export requests data

danilo996:


src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -0,0 +1,105 @@
...
+      LocalDate endDate) {
+    String absenceName = assignAbsenceName(absenceType);
+    String dateRange = assignDateRange(startDate, endDate);
+
+    return absenceName + dateRange + FILE_SUFFIX;
+  }
+
+  private String assignDateRange(LocalDate startDate, LocalDate endDate) {
+    if (startDate != null) {
+      return ":" + startDate.toString() + ":" + endDate.toString();
```
JonathanTolic: We should use - rather than :

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -0,0 +1,105 @@
...
+    var requests = requestRepository.findAllRequestsForExport(absenceTypes, teamId, startDate, endDate,
+        absenceTypesEmpty);
+    setReason(requests);
+
+    return requests;
+  }
+
+  private void setReason(List<Request> requests) {
+    requests.stream().filter(request -> request.getReason().contains(",")).forEach(request -> {
+      var reason = request.getReason().replace(",", "");
```
JonathanTolic: Could we escape the commas rathar than omit them?

src/main/java/eu/execom/hawaii/util/WriteCsvToResponse.java
```diff
@@ -0,0 +1,71 @@
...
+          writer).withMappingStrategy(mappingStrategy)
+                 .withQuotechar(CSVWriter.NO_QUOTE_CHARACTER)
+                 .withSeparator(CSVWriter.DEFAULT_SEPARATOR)
+                 .build();
+
+      csvWriter.writeNext(headers);
+      mappingStrategy.setColumnMapping(columnMapping);
+      beanToCsv.write(requests);
+
+    } catch (Exception e) {
```
JonathanTolic: I am not sure what in this code could throw an exception but try to always catch exceptions as specifically as possible

src/main/java/eu/execom/hawaii/api/controller/DataExportController.java
```diff
@@ -0,0 +1,59 @@
...
+      throws IOException {
+
+    var fileName = dataExportService.assignAdequateNameToRequestsExportFile(absenceType, startDate, endDate);
+    response.setContentType("text/csv; charset=UTF-8");
+    response.setHeader(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"" + fileName + "\"");
+
+    var requests = dataExportService.getRequestForExport(absenceType, teamId, startDate, endDate);
+    var requestDtos = getRequestExportDtos(requests);
+
+    WriteCsvToResponse.writeRequests(response.getWriter(), requestDtos, absenceType);
```
JonathanTolic: if writeRequests is catching its exceptions do we need exportRequest to throw IOExceptions ?

src/main/java/eu/execom/hawaii/restfactory/RequestExportDtoFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+  private int getHoursInRange(Request request) {
+    return allowanceService.calculateHours(request.getDays());
+  }
+
+  private double getDaysInRange(Request request) {
+    return (double) allowanceService.calculateHours(request.getDays()) / 8;
+  }
+
+  public int calculateTotalRequestHours(Long requestId) {
```
JonathanTolic: think we can make this private

src/main/java/eu/execom/hawaii/model/enumerations/UserStatusType.java
```diff
@@ -1,5 +1,15 @@
 package eu.execom.hawaii.model.enumerations;
 
+import lombok.Getter;
+
 public enum UserStatusType {
-  ACTIVE, INACTIVE, DELETED
+  ACTIVE("active"), INACTIVE("inactive"), DELETED("deleted");
```
JonathanTolic: change formatting

src/main/java/eu/execom/hawaii/util/WriteCsvToResponse.java
```diff
@@ -0,0 +1,71 @@
...
+      System.out.println("Writing CSV error!");
+      e.printStackTrace();
+    }
+  }
+
+  private static String[] getHeaders(AbsenceType absenceType) {
+    if (absenceType != null && absenceType.equals(AbsenceType.BONUS)) {
```
stanislavgrujic: You can avoid the null check when you compare the enums the oposite way:
```
if (AbsenceType.BONUS.equals(absenceType)) {
```

src/main/java/eu/execom/hawaii/util/WriteCsvToResponse.java
```diff
@@ -0,0 +1,71 @@
...
+      return BONUS_REQUEST_HEADERS;
+    } else {
+      return LEAVE_REQUEST_HEADERS;
+    }
+  }
+
+  private static String[] getColumnMapping(AbsenceType absenceType) {
+    if (absenceType != null && absenceType.equals(AbsenceType.BONUS)) {
```
stanislavgrujic: You can avoid the null check when you compare the enums the oposite way:
```
if (AbsenceType.BONUS.equals(absenceType)) {
```

src/main/java/eu/execom/hawaii/dto/export/RequestExportDto.java
```diff
@@ -0,0 +1,54 @@
...
+
+  private String reason;
+
+  private int hours;
+
+  private int hoursInRange;
+
+  private double days;
+
+  private double daysInRange;
```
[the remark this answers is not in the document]
danilo996: Because of half day values.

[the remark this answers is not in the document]
stanislavgrujic: Why not use float? You don't need double precision.

PR #640, opened 2019-07-10
B feature/gh 618 export requests data

danilo996:


src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -0,0 +1,124 @@
...
+  private List<AbsenceType> setAbsenceTypes(AbsenceType absenceType) {
+    List<AbsenceType> absenceTypes = new ArrayList<>(List.of(absenceType));
+    if (absenceType.equals(AbsenceType.LEAVE)) {
+      absenceTypes.add(AbsenceType.TRAINING);
+    }
+
+    return absenceTypes;
+  }
+
+  public String assignAdequateNameToRequestsExportFile(AbsenceType absenceType, Long teamId, LocalDate startDate,
```
JonathanTolic: technically we do not assign a name to a file here we just get it. "Adequate" feels somehow wrong... My suggestion would be createExportFileName, or createrequestExportFileName

src/main/java/eu/execom/hawaii/api/controller/DataExportController.java
```diff
@@ -0,0 +1,76 @@
...
+      @RequestParam(required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) LocalDate startDate,
+      @RequestParam(required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) LocalDate endDate)
+      throws IOException {
+
+    var fileName = dataExportService.assignAdequateNameToRequestsExportFile(absenceType, teamId, startDate, endDate);
+    response.setContentType("text/csv; charset=UTF-8");
+    response.setHeader(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"" + fileName + "\"");
+
+    var requests = dataExportService.getRequestForExport(absenceType, teamId, startDate, endDate);
+    var requestDtos = getRequestExportDtos(requests);
```
JonathanTolic: I would completely remove RequestExportDto. RequestExportDtoFactory could be change to RequestCSVExportHelper and have it assemble a string directly. I would not use CSVWriter at all. Reason being we are creating one fewer Dto class, a class that is the only one in an export category. More importantly we loop through requests once to map them to a dto then again to convert them to a string. We can skip one iteration here.

PR #668, opened 2019-07-25
B enhancement/gh-669-allowance import create year if not exist

danilo996:


danilo996:
#669

src/main/java/eu/execom/hawaii/batch/AllowanceProcessor.java
```diff
@@ -98,11 +93,31 @@ private int assignValue(double importedHours) {
...
+  private int calculateNegativeValue(double importedHours) {
+    return calculateHours(Math.abs(importedHours)) * -1;
+  }
+
+  private double roundToLowerValue(double importedHours, double modulus) {
+    return importedHours - modulus;
+  }
+
+  private double roundToHigherValue(double importedHours, double modulus) {
+    return (importedHours + HALF_DAY) - modulus;
```
mariobasic: Please remove brackets.

src/main/java/eu/execom/hawaii/batch/AllowanceProcessor.java
```diff
@@ -98,11 +93,31 @@ private int assignValue(double importedHours) {
...
-    var result = modulus < 2.5 ? importedHours - modulus : (importedHours + HALF_DAY) - modulus;
+    double modulus = importedHours % HALF_DAY;
+    double result = modulus < MODULUS_BOUNDARY ?
+        roundToLowerValue(importedHours, modulus) :
+        roundToHigherValue(importedHours, modulus);
 
     return (int) Math.round(result);
   }
 
+  private int calculateNegativeValue(double importedHours) {
```
mariobasic: This method should be above `calculateHours(double importedHours)` method above.

src/main/java/eu/execom/hawaii/batch/AllowanceProcessor.java
```diff
@@ -98,11 +93,31 @@ private int assignValue(double importedHours) {
    * client, so there is no point in having excess amount of hours at users disposal, when user
    * can't use it. Users allowance will be or slightly increased or slightly decreased.
    */
+  private int assignValue(double importedHours) {
+    double modulus = importedHours % HALF_DAY;
+    int calculatedHours = importedHours < 0 ? calculateNegativeValue(importedHours) : calculateHours(importedHours);
+
+    return modulus == 0 ? (int) importedHours : calculatedHours;
```
mariobasic: If `modulus` is `0` we are returning `importedHours`, that means that calculation of `calculatedHours` at line 98 was unnecessary. Suggestion: 
```
    if (importedHours % HALF_DAY == 0) {
       return (int) importedHours;
    } else {
      return importedHours < 0 ? calculateNegativeValue(importedHours) : calculateHours(importedHours);
    }
```

Issue #637, opened 2019-07-10
Add support for audit information


danilo996:
Frontend should provide support for user, team, and allowance manual adjust audit information.

ffpetrovic:
Discuss if old value should even be shown, since it would only be useful in the first change audit (or not even there when the "Created User" audit bug gets fixed) and the "old value" is really just the "new value" from the previous audit.

PR #701, opened 2019-07-30
F enhancement/audits

ffpetrovic:
Resolves: #637


src/main/frontend/src/components/audit/User/EmployeeAudit.js
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
+import moment from 'moment';
+import getReadableChangeLabel from '../helpers/getReadableChangeLabel';
+import getChangedField from '../helpers/getChangedField';
+
+export default class EmployeeAudit extends Component {
```
aberonja: Stateless component

src/main/frontend/src/components/audit/User/EmployeeAudit.js
```diff
@@ -0,0 +1,39 @@
...
+  render() {
+    const { audit } = this.props;
+
+    return (
+      <div
+        key={audit.modifiedDateTime}
```
aberonja: Is there anything better to use than modifiedDateTime? It seems error prone.

src/main/frontend/src/components/audit/User/EmployeeAudits.js
```diff
@@ -0,0 +1,47 @@
...
+
+class EmployeeAudits extends Component {
+  componentDidMount() {
+    this.props.requestEmployeeAudit(this.props.id);
+  }
+
+  render() {
+    const { audit } = this.props;
+
+    if (audit === null) return null;
```
aberonja: I assume that this can be [] so that's why you checked it this way?

src/main/frontend/src/components/audit/User/EmployeeAudits.js
```diff
@@ -0,0 +1,47 @@
...
+
+    return (
+      <div>
+        <div className="my-2">Employee Audits</div>
+        {audit.map(a => (
```
aberonja: Please use something more descriptive than 'a'

src/main/frontend/src/components/audit/allowance/AllowanceAudit.js
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
+import moment from 'moment';
+import getReadableChangeLabel from '../helpers/getReadableChangeLabel';
+import getChangedField from '../helpers/getChangedField';
+
+export default class AllowanceAudit extends Component {
```
aberonja: Stateless component

src/main/frontend/src/components/audit/allowance/AllowanceAudit.js
```diff
@@ -0,0 +1,39 @@
...
+  render() {
+    const { audit } = this.props;
+
+    return (
+      <div
+        key={audit.modifiedDateTime}
```
aberonja: Same as above

src/main/frontend/src/components/audit/helpers/getChangedField.js
```diff
@@ -0,0 +1,30 @@
+import store from '../../../store/store.js';
+import moment from 'moment';
+import _find from 'lodash/find';
+import { getTeams, getLeaveProfiles } from '../../../store/selectors.js';
+
+export default (field, payload) => {
+  if (payload === null || payload === '') return 'Empty';
```
aberonja: /n

src/main/frontend/src/components/audit/allowance/AllowanceAudit.js
```diff
@@ -0,0 +1,39 @@
...
+        className="d-flex justify-content-between py-2 border-bottom"
+      >
+        <div className="d-flex flex-column">
+          {audit.changesPerformed.map(change => (
+            <div className="mb-2" key={change.fieldName}>
+              <div className="small text-uppercase font-weight-bold mb-1">
```
aberonja: Put text in span, p, h, anything but div.

src/main/frontend/src/components/audit/User/EmployeeAudit.js
ffpetrovic: I was wrong, it's not a unix epoch timestamp, but there's also no ID provided.

src/main/frontend/src/components/audit/User/EmployeeAudits.js
ffpetrovic: Yes, the audit object can be an empty array if the audit is empty (no history).

ffpetrovic:
> A lot of audit files look the same to me, can we do something about this?

We should confirm that all audits will look and behave the same.

src/main/frontend/src/components/audit/helpers/getChangedField.js
ffpetrovic: Also try to come up with a more descriptive name.

Issue #733, opened 2019-08-02
Permission to see requests


mariobasic:
Investigate method `findAllbyUserForYear()` when resolving merge conflicts I noticed that changes have been made in that way to exclude check if `authUser` has permission to access users requests.

PR #745, opened 2019-08-08
b-fix/accessing-requests-for-user-without-permission

mariobasic:
Check if  `authUser` has permission to access requests for user.

Resolves: #733


src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -138,19 +138,17 @@ public UserService(UserRepository userRepository, UserPushTokensRepository userP
...
-  public User getUserById(Long id, User authUser) {
-    return Optional.ofNullable(id).isEmpty() ?
-        getUserById(authUser.getId()) :
-        getUserIfAuthUserHasPermission(id, authUser);
+  public User getUserById(Long userId, User authUser) {
+    return Optional.ofNullable(userId).isEmpty() ? authUser : getUserIfAuthUserHasPermission(userId, authUser);
   }
 
-  private User getUserIfAuthUserHasPermission(Long id, User authUser) {
+  User getUserIfAuthUserHasPermission(Long id, User authUser) {
```
stanislavgrujic: In general it is a bad practice to name methods with some condition stated. It does not make sense to return a user if auth user does not have a condition, so it can be implicitly concluded that the auth user has to have a permission, in order to get the specific user. Please rename to `getUser`.

PR #756, opened 2019-08-13
b-fix/fix-broken-tests

mariobasic:


src/main/java/eu/execom/hawaii/security/IdTokenVerifierFilter.java
```diff
@@ -22,7 +22,7 @@
 
 @Slf4j
 public class IdTokenVerifierFilter extends OncePerRequestFilter {
-  public static final String ID_TOKEN_HEADER = "X-ID-TOKEN";
+  static final String ID_TOKEN_HEADER = "X-ID-TOKEN";
```
stanislavgrujic: Can you make this private? Is it used anywhere else?

Issue #718, opened 2019-08-02
Create public holidays for newly created year


danilo996:
- we should implement business logic which would handle creation of public holidays for newly created year. We should create only holidays for which we are sure that will happen on given date. (for e.g. Easter is falling every year at different time so it will be created manually)

- also, if public holiday falls on Sunday, we would create it on following Monday.

- public holidays which will be created automatically: New Year(1st and 2nd January), Christmas(7th January), Serbian statehood day(15th February), Workers day(1st and 2nd May),  Armistice Day in World War One(11th November) ;

PR #765, opened 2019-08-19
b-feature/gh-718-create-public-holidays-for-newly-created-year

danilo996:
Automatically create public holidays when new year has been created

- if holiday date (except Christmas) falls on Sunday, it will be persisted in database with date set on following day(Monday)

- if holiday consists of 2 days and first day falls on Sunday, 1st and 2nd days of holiday will be changed to following day


stanislavgrujic:
This implementation is wrong in several ways:
1. It does not cover all public holidays (e.g., Easter is missing)
2. It does not allow multitenancy, where tenant other than Execom is in another country which has different public holidays.

My suggestion is to create an import function which allows a user to import these dates in the application by sending CSV with:
```
<date>,<holiday_name>
```

This way it will be completely flexible and easy for users to add public holidays for a new year. 
Other option is to add in UI an option to copy holidays from previous year, with additional flag in public holiday configuration if it is moving holiday, such as Easter is.

src/main/java/eu/execom/hawaii/service/PublicHolidayService.java
```diff
@@ -87,4 +93,64 @@ public void delete(Long id) {
...
+        PublicHolidaysInformationProvider.STATEHOOD_DAY);
+    var statehoodDaySecond = createPublicHoliday(year, PublicHolidaysInformationProvider.STATEHOOD_DAY_DATE_2ND_DATE,
+        PublicHolidaysInformationProvider.STATEHOOD_DAY);
+    var workersDay = createPublicHoliday(year, PublicHolidaysInformationProvider.WORKERS_DAY_FIRST_DATE,
+        PublicHolidaysInformationProvider.WORKERS_DAY);
+    var secondWorkersDay = createPublicHoliday(year, PublicHolidaysInformationProvider.WORKERS_DAY_SECOND_DATE,
+        PublicHolidaysInformationProvider.WORKERS_DAY);
+    var armisticeDay = createPublicHoliday(year, PublicHolidaysInformationProvider.ARMISTICE_DAY_DATE,
+        PublicHolidaysInformationProvider.ARMISTICE_DAY);
+    List<PublicHoliday> newPublicHolidays = List.of(newYearFirstDay, newYearSecondDay, christmas, statehoodDay,
```
stanislavgrujic: Please see my comment about hard coding public holidays logic in code.

danilo996: We agreed with our clients that these holidays only will be created when new year has been created. As for other two things, you are absolutely right. This logic is not extensible at all, not to mention multitenancy. In my opinion, your second suggestion is better than first. It will be very easy to use looking from users perspective. Maybe we should copy all public holidays from previous year, when creating new year, and leave it to end user to choose which public holidays should be auto created. Then i would apply my logic to public holidays he has chosen.  I will close this pull request and definitely open discussion with other colleagues about this.

Issue #794, opened 2019-08-27
fix afternoon only issue


BraneOpacic21:

BraneOpacic21:
![afternoonOnly](https://user-images.githubusercontent.com/45173769/63762602-67f81400-c8c3-11e9-9472-aec059997dbf.gif)

PR #802, opened 2019-08-27
Fix request day duration

ffpetrovic:
This commit fixes the issue when creating a new leave request with a
"afternoon only first day" and/or "morning only last day".

See: #794


aberonja:
Please, name your branch using this kind of naming
`f-fix/gh-777-request-duration`

PR #804, opened 2019-08-29
F enhancement/gh 800 email buttons

ffpetrovic:


src/main/frontend/src/components/common/InputChips.js
```diff
@@ -0,0 +1,91 @@
+import React, { Component, Fragment } from 'react';
+import { Chip } from '@material-ui/core';
+import styled from 'styled-components';
+
+const ChipsWrapper = styled.div`
+  display: flex;
```
aberonja: Use bootstrap classes for this, no need for a component.

src/main/frontend/src/components/common/InputChips.js
```diff
@@ -0,0 +1,91 @@
...
+  justify-content: flex-start;
+  flex-wrap: wrap;
+`;
+
+const BlankInput = styled.input`
+  border: none;
+  background: transparent;
+  display: inline-block;
+  flex: 1 1 auto;
+  padding: 0.25rem;
```
aberonja: Almost all of this can be bootstrap class

src/main/frontend/src/components/common/InputChips.js
```diff
@@ -0,0 +1,91 @@
...
+`;
+
+const ChipWrapper = styled.div`
+  margin: 2px 2px 2px 0;
+  .MuiChip-root {
```
aberonja: /n

src/main/frontend/src/components/common/InputChips.js
```diff
@@ -0,0 +1,91 @@
...
+    height: 23px;
+    .MuiChip-deleteIcon {
```
aberonja: /n

src/main/frontend/src/components/common/InputChips.js
```diff
@@ -0,0 +1,91 @@
...
+    const sanitizedChip = this.sanitize(chip);
+
+    if (!chips.includes(sanitizedChip) && /\S/.test(chip)) {
+      this.setState({ chips: [...chips, sanitizedChip] }, this.handleChange);
+    }
+  };
+
+  removeChip = i => {
+    const { chips } = this.state;
+    this.setState(
```
aberonja: /n

aberonja:
Also, validate emails before converting them to chips.

ffpetrovic:
> Also, validate emails before converting them to chips.

Validate emails or prefill as emails?

aberonja:
Both, because this is going to be invalid `@!e93w0932108dasda!=+_)_)(_)         @execom.eu`, right?

ffpetrovic:
> Both, because this is going to be invalid `@!e93w0932108dasda!=+_)_)(_) @execom.eu`, right?

How do I differentiate between an invalid email and just an email name without a domain (which would also be considered an invalid email)? Contains `@`?

aberonja:
You can prefill first if needed, and then validate email.

aberonja:
It works fine now. Check with the rest of the team about using tab on keyboard to make a chip and enter as well. Also check with them about string removal when it's already used.

Issue #832, opened 2019-09-05
Edit public holiday error


aberonja:
Please, change date formatting to dd MMM yyyy (05 Sep 2019).

![image](https://user-images.githubusercontent.com/35724583/64331264-c3688700-cfd2-11e9-8a9f-48b6dc649246.png)

mariobasic:
![image](https://user-images.githubusercontent.com/45662818/64333296-8f8f6080-cfd6-11e9-85a0-347376486760.png)

mariobasic:
This format is the one requested?

aberonja:
Can you make it without any dashes? 
07 May 2019

PR #835, opened 2019-09-05
b-fix/public-holiday

mariobasic:
Ordering of public holidays to be by date not name.
Closes: #833

Exception message date format change.
Closes: #832


src/main/java/eu/execom/hawaii/service/PublicHolidayService.java
```diff
@@ -76,7 +77,8 @@ public PublicHoliday getById(Long id) {
   public PublicHoliday save(PublicHoliday publicHoliday) {
     if (publicHolidayRepository.existsByDate(publicHoliday.getDate())) {
       throw new DataIntegrityViolationException(
-          "Public holiday with date : " + publicHoliday.getDate() + " already exists.");
+          "Public holiday with date : " + publicHoliday.getDate().format(DateTimeFormatter.ofPattern("dd MMM yyyy"))
```
stanislavgrujic: It is very inefficient to instantiate the DateTimeFormatter each time there is an exception, please extract it to a constant with a meaningful name. Also, we already discussed about chaining method calls - it is difficult to read and difficult to debug, please avoid doing it.

PR #836, opened 2019-09-05
b-fix/gh-833-public-holiday-exists-by-date-validation-when-updating

danilo996:


src/main/java/eu/execom/hawaii/service/PublicHolidayService.java
```diff
@@ -88,6 +88,28 @@ public PublicHoliday save(PublicHoliday publicHoliday) {
     return publicHoliday;
   }
 
+  public PublicHoliday update(PublicHoliday publicHoliday) {
+    validateDateIsUnique(publicHoliday);
+    try {
+      publicHolidayRepository.save(publicHoliday);
+    } catch (DataIntegrityViolationException e) {
+      throw new DataIntegrityViolationException("Public holiday name too long.");
```
stanislavgrujic: This is wrong because:
- you are **assuming** that the `DataIntegrityViolationException` is thrown because the holiday name is too long. This may be accurate at the moment, but can change in the future.
- you have let pass data all the way to the database even though you could have catch it by validating data received from the client.

Please remove this and validate the size of the received public holiday name  in controller (Add `@Valid` annotation on PublicHolidayDto and add maximum name size).

src/main/java/eu/execom/hawaii/service/PublicHolidayService.java
```diff
@@ -88,6 +88,28 @@ public PublicHoliday save(PublicHoliday publicHoliday) {
...
+    }
+
+    return publicHoliday;
+  }
+
+  private void validateDateIsUnique(PublicHoliday publicHoliday) {
+    var oldPublicHoliday = publicHolidayRepository.getOne(publicHoliday.getId());
+    boolean dateUpdated = !oldPublicHoliday.getDate().equals(publicHoliday.getDate());
+    boolean existsInDatabase = publicHolidayRepository.existsByDate(publicHoliday.getDate());
```
stanislavgrujic: How do you determine if you will use `var` or a concrete type? Please be consistent with type declaration.

stanislavgrujic: Still not resolved.

danilo996: I use var when the variable type can be clearly determined just by reading its name. Also, I use var when all of the variables in method body are of the same type.

danilo996: Should I use `var` for all three variables?

stanislavgrujic: My point is that it should not be mixed, or at least that you should have very clear rules on when to put type and when to infer it. Just think about it, this is OK as it is.

PR #846, opened 2019-09-06
b-fix/data-integrity-violation-checks

mariobasic:
Validate data received from the client 

Resolves: #839 
Resolves: #840
Resolves: #841 
Resolves: #842
Resolves: #843
Resolves: #845


src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -7,88 +7,108 @@
...
-    return new ResponseEntity<>(Collections.singletonMap("error", ex.getMessage()),
+    return new ResponseEntity<>(Collections.singletonMap(ERROR, ex.getMessage()),
         HttpStatus.REQUESTED_RANGE_NOT_SATISFIABLE);
   }
 
   @ExceptionHandler(Exception.class)
   public ResponseEntity<Map<String, String>> handle(HttpServletRequest request, Exception ex) {
     logException(request, ex);
-    return new ResponseEntity<>(Collections.singletonMap("error", ex.getMessage()), HttpStatus.INTERNAL_SERVER_ERROR);
+    return new ResponseEntity<>(Collections.singletonMap(ERROR, ex.getMessage()), HttpStatus.INTERNAL_SERVER_ERROR);
```
stanislavgrujic: You could extract a method and replace all duplications by
```
private static ResponseEntity<Map<String, String>> errorResponse(String message, HttpStatus statusCode) {
  return new ResponseEntity<>(Collections.singletonMap(ERROR, message), statusCode);
}
```

The usage would be as simple as:
```
return errorResponse(ex.getMessage(), HttpStatus.INTERNAL_SERVER_ERROR);
```

src/main/java/eu/execom/hawaii/dto/RequestDto.java
```diff
@@ -25,12 +26,16 @@
   private RequestStatus requestStatus;
 
   @NotNull
+  @Size(max = 255, message = "Reason is too long.")
   private String reason;
 
+  @Size(max = 255, message = "Comment is too long.")
   private String approverComment;
 
+  @Size(max = 255, message = "Reason is too long.")
```
stanislavgrujic: Maybe use *Cancellation reason is too long.* to differ it from the request reason?

src/main/java/eu/execom/hawaii/dto/RequestDto.java
```diff
@@ -25,12 +26,16 @@
...
   private String cancellationReason;
 
+  @Size(max = 255, message = "Comment is too long.")
```
stanislavgrujic: Maybe use *Cancellation comment is too long.* to differ it from the request comment?

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -7,91 +7,130 @@
...
+    return errorResponse(Collections.singletonList(ex.getMessage()), HttpStatus.REQUESTED_RANGE_NOT_SATISFIABLE);
   }
 
   @ExceptionHandler(Exception.class)
-  public ResponseEntity<Map<String, String>> handle(HttpServletRequest request, Exception ex) {
+  public ResponseEntity<Map<String, List<String>>> handle(HttpServletRequest request, Exception ex) {
     logException(request, ex);
-    return new ResponseEntity<>(Collections.singletonMap("error", ex.getMessage()), HttpStatus.INTERNAL_SERVER_ERROR);
+
+    return errorResponse(Collections.singletonList(ex.getMessage()), HttpStatus.INTERNAL_SERVER_ERROR);
```
stanislavgrujic: Why do you need a singletonList? It does not help much if you replace calls in all of the methods of `Collection.singletonMap` with a `Collection.singletonList`. Can you internalize it inside the `errorResponse` method??

stanislavgrujic: I see now that you have one method that returns multiple messages. You should not change all your responses/methods for the sake of absolute re-usage within a single method. Rather split it into two. Always think about the balance of code reusability/duplication and readability.

mariobasic: Yes, that's why. I have thought about that, thanks for suggestion/insight, I will change it accordingly.

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -7,91 +7,130 @@
...
-    return new ResponseEntity<>(Collections.singletonMap("error", ex.getMessage()), HttpStatus.INTERNAL_SERVER_ERROR);
+
+    return errorResponse(ex.getMessage(), HttpStatus.INTERNAL_SERVER_ERROR);
   }
 
   private void logException(HttpServletRequest request, Exception ex) {
     log.error("Requesting resource: '{}' raised exception: '{}'", request.getRequestURI(), ex);
   }
+
+  private ResponseEntity<Map<String, String>> errorResponse(String message, HttpStatus badRequest) {
```
stanislavgrujic: Please rename `badRequest` to `statusCode`.

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -7,91 +7,130 @@
...
+    return new ResponseEntity<>(Collections.singletonMap(ERROR, message), badRequest);
+  }
+
+  private ResponseEntity<Map<String, List<String>>> errorResponse(List<String> message) {
+    return new ResponseEntity<>(Collections.singletonMap(ERROR, message), HttpStatus.BAD_REQUEST);
```
stanislavgrujic: Please extract `statusCode` as a method parameter, instead of hard-coding it inside. In case you have another usage of this method, it won't be correct.

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -7,91 +7,130 @@
...
 
   private void logException(HttpServletRequest request, Exception ex) {
     log.error("Requesting resource: '{}' raised exception: '{}'", request.getRequestURI(), ex);
   }
+
+  private ResponseEntity<Map<String, String>> errorResponse(String message, HttpStatus badRequest) {
+    return new ResponseEntity<>(Collections.singletonMap(ERROR, message), badRequest);
+  }
+
+  private ResponseEntity<Map<String, List<String>>> errorResponse(List<String> message) {
```
stanislavgrujic: Please use plural in naming, since it is more than a single message.

PR #851, opened 2019-09-06
f-fix/gh-808-teams-endpoint-forbidden

### LINKED ISSUE #808: User not admin, /teams endpoint forbidden

ffpetrovic:
Closes #808


ffpetrovic:
Working fine, since we only edited the `getTeamsCalendar` and it's `getTeamsCalendarApi` counterpart, it should be just fine, especially since this function/saga is being used only in one place.

aberonja:
You're using the same object in the store for both values. Find a way to separate them somehow, it's your choice ♻️

ffpetrovic:
Derped on that last commit, ended up on the wrong branch. Fixes #854 tho

PR #853, opened 2019-09-09
f-fix/gh-848-public-holidays-same-data-twice

### LINKED ISSUE #848: Public holidays same data twice
image

aberonja:
Closes #848


src/main/frontend/src/pages/administration/components/common/AdministrationTable.js
```diff
@@ -196,7 +194,7 @@ class AdministrationTable extends Component {
                 loadMore={this.loadMore}
                 hasMore={!last}
                 threshold={50}
-                initialLoad={false}
+                initialLoad={initialLoad || false}
```
ffpetrovic: `{initialLoad}`

aberonja: It won't work if I make it like that, trust me 😄
initialLoad

Issue #838, opened 2019-09-05
Team edit members don't update


aberonja:
![team-edit-bug](https://user-images.githubusercontent.com/35724583/64350456-f9b8fd00-cff8-11e9-8df9-f40d2acd0f4e.gif)

PR #862, opened 2019-09-10
b-fix/disable-employee-removal-when-updating-team

mariobasic:
Users must belong to a team so it should be forbidden to remove employees from a team while editing team.

Resolves: #838


src/main/java/eu/execom/hawaii/service/TeamService.java
```diff
@@ -148,6 +151,12 @@ private void validateTeamNameIsUnique(String teamName, String oldTeamName) {
     }
   }
 
+  private void validateThatUsersAreNotRemoved(Team updatedTeam, Team databaseTeam) {
```
stanislavgrujic: Please be consistent in naming. Methods on lines 137 and 138 are named *validate...*, while this method is *validateThat*. Please remove *That* from the method name.

PR #905, opened 2019-09-18
b-fix/dependency-update-produced-bugs-user-service

danilo996:
Resolves: #906 , #908


src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -180,7 +180,9 @@ User getUser(Long id, User authUser) {
    */
   @Transactional(readOnly = true)
   public User getUserById(Long id) {
-    return userRepository.getOne(id);
+    return userRepository.findById(id)
```
mariobasic: Someone reading this method name would think that this method gets user using `getOne()` repository call. Please rename method or better yet create new method named `findById` and use this repository call there.

Also there is a lot of method invocations for this method `public User getUserById(Long id)` maybe for some of them wanted behavior is to make repository call using `userRepository.getOne(id)` instead of `userRepository.findById(id)` 🤔

danilo996: I have already checked all invocations. As for naming, you couldn't be more right.

Issue #937, opened 2019-09-23
Same user can be approver and member of the same team


rgerstmajer-ec:
Steps to reproduce:

- Log in as HR Manager
- Click on "Create new eployee" button
- Fill in necessary information
- Assign user as approver of a team
- Assign user as member of same team
- Click save

Expected result: User cannot be created while being member and apporver of same team.
Actual result: User is created with no error messages.

![Same_approver_and_member](https://user-images.githubusercontent.com/54444325/65435650-26567c80-de21-11e9-8fc2-0da34448ec5f.gif)

rgerstmajer-ec:
Note: Side effect of this is that the user gets logged out with error: "You are not authorized to approve/reject this request" when trying to  approve/reject their own request.

Steps to reproduce:

- Log in as user that is both approver and member of the same team.
- Submit a leave request or sickness cancellation request.
- Go to Approvals page
- Select that request
- Click approve or reject

Expected result: The action gets completed or not, redirecting to the appropriate page
Actual result: User gets logged out forcefully.


![Self_approve_error](https://user-images.githubusercontent.com/54444325/65488448-74af5e00-dea9-11e9-803d-12db9ba1d79b.gif)

PR #944, opened 2019-09-24
b-fix/gh-937-do-not-allow-user-to-become-member-and-approver-of-the-same-team

petrovicdanilo:
Resolves: #937


src/test/java/eu/execom/hawaii/service/UserServiceTest.java
```diff
@@ -449,12 +454,14 @@ public void shouldCreateUserWithAllowance() {
     verifyNoMoreInteractions(allMocks);
   }
 
-  @Test(expected = DataIntegrityViolationException.class)
+  @Test(expected = EntityExistsException.class)
   public void shouldFailToCreateUserBecauseEmailExists() {
     // given
+
```
mariobasic: Why did you added empty lines here and below at line 464? 👀

PR #951, opened 2019-09-25
f-fix/styles-on-different-browsers

jprotic:
-Fix borders on select year field on Firefox
 -Fix invisible borders on weekend and non-existent days in dashboard on Firefox
 -Fix invisible borders of days in Team calendar on Firefox
 -Align day number fields on Firefox
 -Fix year and team name fields borders on Firefox


ffpetrovic:
There's a problem with the calendar now.
image

This is how it looked like before:
image

Issue #940, opened 2019-09-24
Create new year should have more strict validations


rgerstmajer-ec:
Steps to reproduce:

- Log in as HR manager
- Go to Administration Page
- Click on "Add year" button ( red plus sign in the top right corner )
- Enter various values in the "Year" field 

Expected result: Some years should not be allowed to be created
Actual results: Any year in the range of -999 to 9999 can be created

![Yeaf_Validation](https://user-images.githubusercontent.com/54444325/65507789-f5348580-dece-11e9-90d5-257f2aad7807.png)

PR #953, opened 2019-09-25
Create and update year validation

petrovicdanilo:
Resolves: #940


src/main/java/eu/execom/hawaii/service/YearService.java
```diff
@@ -27,6 +28,8 @@
   private UserRepository userRepository;
   private AllowanceRepository allowanceRepository;
 
+  private static final String YEAR_REGEX = "^(20|21|22|23|24|25)\\d{2}$";
```
stanislavgrujic: Are we checking the year for the next 5 centuries?

src/main/java/eu/execom/hawaii/service/YearService.java
```diff
@@ -88,4 +94,18 @@ Allowance createAllowance(Year createdYear, User user) {
...
+
+  public void delete(Long id) {
+    var year = findById(id);
+    validateYearIsRemovable(year);
+
+    yearRepository.deleteById(id);
+  }
+
+  private void validateYearIsRemovable(Year year) {
+    if (year.isActive() && !year.getAllowances().isEmpty()) {
```
stanislavgrujic: Is this condition correct? If a year is active, but there are no allowances, it will be removed. Is that the expected behavior?

petrovicdanilo: Yes, active year with no allowances assigned has no relations to other entities so it can be safely removed. However, i could remove second condition because changing year status to active in our application triggers allowance service method which assigns allowance to all active users.

PR #973, opened 2019-10-01
f-fix/allow-negative-manual-adjust-values

ffpetrovic:
- Add error toast to ManualAdjust
 - Allow negative values in ManualAdjust


src/main/frontend/src/pages/administration/components/manual-adjust/ManualAdjust.js
```diff
@@ -89,6 +85,22 @@ class ManualAdjust extends Component {
...
+    } else {
+      return 'Invalid Values';
+    }
+  };
+
+  tryParseInput = n => {
+    if (!isNaN(parseInt(n, 10))) {
+      return parseInt(n, 10);
+    } else {
+      return n;
```
aberonja: Don't use 'n', please write number 🗡

PR #980, opened 2019-10-02
f-fix/public-pathname-for-staging

ffpetrovic:


src/main/frontend/src/env.js
```diff
@@ -1,6 +1,16 @@
...
+  if (process.env.NODE_ENV === 'production' && publicUrl) {
+    const pathname = new URL(publicUrl).pathname;
+
+    if (pathname !== '/') {
+      return pathname;
+    }
+    return '';
+  }
+
+  return '';
```
aberonja: Just testing a comment

Issue #608, opened 2019-07-01
Import from Google to be optional


mariobasic:
If that leave record was already imported to be skipped.

PR #1004, opened 2019-10-08
B-fix/gh-608-make-calendar-request-import-optional

mariobasic:
Resolves: #608


src/main/java/eu/execom/hawaii/util/RequestImportConstants.java
```diff
@@ -0,0 +1,28 @@
...
+
+  public static final String FEVER = "Fever";
+
+  public static final String PENDING = "Pending Leave";
+
+  public static final String DECEASE_FAMILY_MEMBER = "Decease family member";
+
+  public static final String FUNERAL_FAMILY_MEMBER = "Funeral family member";
+
+  public static final String DESEASE_FAMILY_MEMBER = "Desease family member";
```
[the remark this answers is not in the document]
mariobasic: Yes I know, it seems that for some time there was a type of leave in Appogee Leave with this spelling mistake, and that mistake is in import records also, so in order to match requests I need to write it this way.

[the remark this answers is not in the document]
stanislavgrujic: Please add a comment above to make sure than no-one accidentally "fixes" it. :) Something like *Intentional spelling error to match the name from Appogee*. Also, it would help to add a comment above the section marking which constants are used in that case. E.g.,
```
// Leave request reasons from Appogeee
```

src/main/java/eu/execom/hawaii/batch/AppogeeLeaveProcessor.java
```diff
@@ -32,22 +32,29 @@
...
 @Slf4j
 @Component
 public class AppogeeLeaveProcessor implements ItemProcessor<AppogeeLeaveRequestImportData, Request> {
 
-  private static final int FULL_DAY = 8;
-  private static final int HALF_DAY = 4;
   private static final Map<String, RequestStatus> REQUEST_STATUS = new HashMap<>();
   private static List<LocalDate> PUBLIC_HOLIDAY_DATES = new ArrayList<>();
   private static List<Absence> ABSENCES = new ArrayList<>();
+  static Map<String, List<Request>> PERSISTED_REQUESTS = new HashMap<>();
```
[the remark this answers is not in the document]
mariobasic: There is, I could duplicate the logic to populate this field from another class like this: 
```
    List<Request> persistedRequests = requestRepository.findAllApprovedWithDaysEager();
    persistedRequests.forEach(
        request -> PERSISTED_REQUESTS.merge(request.getUser().getEmail(), List.of(request), (requests, newRequest) -> {
          List<Request> updatedRequests = new ArrayList<>(requests);
          updatedRequests.add(request);
          return updatedRequests;
        }));
  }
```
so there will no longer be a need for this original field to be static, but I will have the same two maps in two different classes.

Or I could get all requests for a given user in the time of importing requests which would result in many repository calls but maybe better because it will not happen often.

Should I change code logic to the ether of these options, or is there a better solution that I'm not seeing?

src/main/java/eu/execom/hawaii/batch/RequestProcessor.java
```diff
@@ -122,12 +146,34 @@ private String getRequestStatus(CalendarImportData calendarImportData) {
...
+      absenceName = FEVER;
     } else {
       absenceName = calendarImportData.getEventDescription().split(":")[1].trim();
     }
 
     return appogeeLeaveProcessor.getAbsence(absenceName);
   }
 
+  private boolean isRequestAlreadyImported(CalendarImportData calendarImportData, User user) {
+    return Optional.ofNullable(appogeeLeaveProcessor.PERSISTED_REQUESTS.get(user.getEmail()))
```
stanislavgrujic: You are exposing internal structure of `AppogeeLeaveProcessor` by making the `PERSISTED_REQUESTS` available to the rest of the package. It is a bad design, since it allows users of your class also to make changes to this instance. What you should do instead is to create a method in `AppogeeLeaveProcessor` which would check/obtain from this field what you need. E.g. (in `AppogeeLeaveProcessor` class),
```
List<Request> getPersistedRequestsForUser(User user) {
  return PERSISTED_REQUESTS.getOrDefault(user.getEmail(), Collections.emptyList());
}
```
Also, you might need to think about synchronisation of this field since it may be used from two different threads at the same time (not saying you have to do it, just think about it and apply, if necessary :) ).
I think the rest of the code should be simpler since you're not using Optional for null value, but an empty list.

src/main/java/eu/execom/hawaii/batch/AppogeeLeaveProcessor.java
```diff
@@ -253,7 +260,12 @@ private Request createRequestFromImport(AppogeeLeaveRequestImportData leaveImpor
   }
 
   Optional<Absence> getAbsence(String absenceName) {
-    return ABSENCES.stream().filter(absence -> absence.getName().equals(absenceName)).findFirst();
+    absenceName = absenceName.equals(DECEASE_FAMILY_MEMBER) ? FUNERAL_FAMILY_MEMBER : absenceName;
+    absenceName = absenceName.equals(DESEASE_FAMILY_MEMBER) ? SERIOUS_ILLNESS_OF_CLOSE_FAMILY_MEMBER : absenceName;
+    absenceName = absenceName.equals(WEDDING_OF_FAMILY_MEMBER) ? FAMILY_MEMBER_WEDDING : absenceName;
```
stanislavgrujic: In cases when you compare a constant with a variable value, it's better if you revert the equals checks, since you automatically make the code null safe. E.g,
```
absenceName = DECEASE_FAMILY_MEMBER.equals(absenceName) ? FUNERAL_FAMILY_MEMBER : absenceName
```

src/main/java/eu/execom/hawaii/batch/RequestProcessor.java
```diff
@@ -91,28 +121,24 @@ private Duration getDuration(CalendarImportData calendarImportData, LocalDate cu
     }
   }
 
-  private Request createRequestFromImport(CalendarImportData calendarImportData, List<Day> days) {
-    Optional<User> user = userRepository.findFirstByFullName(getUsersFullName(calendarImportData));
-    if (user.isEmpty()) {
+  private Request createRequestFromImport(CalendarImportData calendarImportData, List<Day> days, User user) {
+    String requestStatus = getRequestStatus(calendarImportData);
+    if (requestStatus.equals(PENDING) || days.isEmpty()) {
```
stanislavgrujic: Same as above - you can make the code null safe if you reverse the equals check.

src/main/java/eu/execom/hawaii/batch/RequestProcessor.java
```diff
@@ -122,12 +148,35 @@ private String getRequestStatus(CalendarImportData calendarImportData) {
...
       absenceName = calendarImportData.getEventDescription().split(":")[1].trim();
     }
 
     return appogeeLeaveProcessor.getAbsence(absenceName);
   }
 
+  private boolean isRequestAlreadyImported(CalendarImportData calendarImportData, User user) {
+    var persistedRequests = appogeeLeaveProcessor.getPersistedRequestsForUser(user);
+
+    return !persistedRequests.equals(Collections.emptyList()) && isRequestAlreadyImported(calendarImportData,
```
stanislavgrujic: Why do you have to compare it with an empty list? Why not just use `!persistedRequests.isEmpty()`

src/main/java/eu/execom/hawaii/service/EmailService.java
```diff
@@ -100,22 +100,33 @@ public void sendEmailForApproval(Request request) {
...
         EmailTemplateProvider.LEAVE_PROFILE_UPDATE);
   }
 
+  /**
+   * Create email to send notification to hr managers that requests for uses with the same name may be incorrect due
+   * to skipped import from Google Calendar.
+   *
+   * @param user the User.
+   */
+  public void sendRequestImportOmittedDueToSameFullName(User user) {
```
stanislavgrujic: Make it Async.

stanislavgrujic:

Issue #1024, opened 2019-10-24
Years not sorted in Administration menu


rgerstmajer-ec:
Steps to reproduce:

- Log in as HR Manager
- Go to Administration menu
- Select Years tab
- Add 3 or more new years in random order ( e.g. 2019, 2020, 2018 )

Expected result: Years are sorted in descending order
Actual result: Years are not sorted

![image](https://user-images.githubusercontent.com/54444325/67486637-aebc7d00-f66c-11e9-8459-4454fb01e3b5.png)

PR #1031, opened 2019-10-29
f-fix/gh-1024-sort-years-in-descending-order

jprotic:
Resolves: #1024


src/main/frontend/src/pages/administration/components/years/Years.js
```diff
@@ -18,11 +18,12 @@ class Teams extends Component {
   render() {
     const { last, results, page } = this.props.years;
 
-    const yearItems = results.length
-      ? results.map(item => {
-          return <YearItem key={item.id} year={item} />;
-        })
-      : null;
+    const yearItems =
+      results.length && results.sort((a, b) => b.year - a.year)
```
iscekic: `sort` mutates an array, let's avoid that

iscekic: Let's also make this more readable - use an `if`. 💃

src/main/frontend/src/pages/administration/components/years/Years.js
```diff
@@ -17,18 +17,20 @@ class Teams extends Component {
...
+    const sortedResults = [...results].sort((a, b) => b.year - a.year);
+    const yearItems = null;
 
-    const yearItems = results.length
-      ? results.map(item => {
-          return <YearItem key={item.id} year={item} />;
-        })
-      : null;
+    if (sortedResults.length) {
+      this.yearItems = sortedResults.map(item => {
```
iscekic: We have both `yearItems` and `this.yearItems` here - we don't want both. Why not just:

```js
const sortedResults = [...results].sort((a, b) => b.year - a.year);
const yearItems = sortedResults.map(/* ... */);

if (!yearItems.length) return null;

return (/* ... */);
```

Issue #1014, opened 2019-10-17
Represent time in day/s metrics


petarkovac:
The suggestion is for the frontend to be responsible for converting hourly metrics, sent from the backend, to days.

PR #1032, opened 2019-10-29
f-fix/gh-1014-represent-time-in-days

jprotic:
Resolves: #1014


src/main/frontend/src/components/audit/helpers/getReadableChangeValue.js
```diff
@@ -33,13 +33,14 @@ export default (field, payload) => {
...
     case 'manualAdjust':
-      return `${payload} hours`;
+      return `${payload / 8} ${payload / 8 === 1 ? 'day' : 'days'}`;
+
     case 'trainingManualAdjust':
-      return `${payload} hours`;
+      return `${payload / 8} ${payload / 8 === 1 ? 'day' : 'days'}`;
     case 'bonusManualAdjust':
-      return `${payload} hours`;
+      return `${payload / 8} ${payload / 8 === 1 ? 'day' : 'days'}`;
```
iscekic: This can potentially show lots of decimals, lets somehow make it only show 2. 😄

iscekic: Similar with all the other places where we divide - multiplication is ok.

src/main/frontend/src/components/audit/helpers/getReadableChangeValue.js
```diff
@@ -33,14 +34,14 @@ export default (field, payload) => {
     case 'comment':
       return payload;
     case 'manualAdjust':
-      return `${payload / 8} ${payload / 8 === 1 ? 'day' : 'days'}`;
+      return `${number(payload / 8)} ${payload / 8 === 1 ? 'day' : 'days'}`;
```
iscekic: Let's make this shorter:

```js
// example
case 'xx':
case 'yy':
  return payload;
case 'zz':
case 'bb':
case 'cc':
  return whatever;
```

src/main/frontend/src/pages/administration/helpers.js
```diff
@@ -22,3 +22,6 @@ export const calculatePage = (employeesTab, employeePage, page) => {
 
   return employeePage + 1;
 };
+
+export const number = value =>
```
iscekic: Interesting function, let's name it better.

src/main/frontend/src/store/sagas/leaveProfileSaga.js
```diff
@@ -34,6 +34,13 @@ import { setShouldTableScrollToTop } from '../actions/administrationActions';
 const getLeaveProfileSaga = function*({ payload }) {
   const leaveProfile = yield call(getLeaveProfileApi(payload));
 
+  leaveProfile.entitlement = leaveProfile.entitlement / 8;
```
iscekic: Let's create copies instead of mutating (throughout this file).

src/main/frontend/src/pages/administration/components/manual-adjust/ManualAdjust.js
```diff
@@ -392,28 +401,21 @@ class ManualAdjust extends Component {
...
-                                this.tryParse(values.bonusManualAdjust) +
-                                this.tryParse(values.carriedOver) +
-                                this.tryParse(values.manualAdjust) -
-                                this.tryParse(takenAnnual) -
-                                this.tryParse(pendingAnnual)
-                            ) / 8
-                          }
+                          value={number(
+                            annual +
+                              Number(values.bonusManualAdjust) +
```
iscekic: This can fail if we pass a non-number, eg. `undefined`. Or did you mean to use `number`?

src/main/frontend/src/pages/administration/components/manual-adjust/ManualAdjust.js
```diff
@@ -61,26 +70,44 @@ class ManualAdjust extends Component {
...
+      obj.carriedOver = Number(obj.carriedOver) * 8;
+      obj.manualAdjust = Number(obj.manualAdjust) * 8;
+      obj.trainingManualAdjust = Number(obj.trainingManualAdjust) * 8;
+
+      return obj;
+    };
+
+    if (!allowance) return null;
+
+    allowance.annual = allowance.annual / 8;
```
iscekic: Multiple things I dislike about this:

* Mutating props is an offence punishable by law 😄 
* Inline function defined on each render
* Lots of repetition

If this component needs data to be in a certain shape (eg. in this case divided by 8), either do it in the component above or in a lifecycle hook.

For the repetition, since we're doing the same thing but with different property names, we can make this much cleaner by abusing javascript stringly-typed-ness. 💃

src/main/frontend/src/pages/administration/components/manual-adjust/ManualAdjust.js
jprotic: Yes, that field is only for numbers.

src/main/frontend/src/pages/administration/components/manual-adjust/ManualAdjust.js
```diff
@@ -83,28 +68,46 @@ class ManualAdjust extends Component {
...
+      'carriedOver',
+      'manualAdjust',
+      'trainingManualAdjust',
+      'pendingTraining',
+      'pendingAnnual',
+      'takenAnnual',
+      'takenTraining'
+    ];
+
+    typeOfAllowance.forEach(item => (allowance[item] = allowance[item] / 8));
```
iscekic: Let's not modify the argument.

src/main/frontend/src/pages/administration/components/manual-adjust/ManualAdjust.js
```diff
@@ -83,28 +68,46 @@ class ManualAdjust extends Component {
...
+
+    return allowance;
   };
 
+  closeDrawer = () => this.props.closeDrawer();
```
iscekic: Why is this needed instead of calling `this.props.closeDrawer` directly?

src/main/frontend/src/pages/administration/components/manual-adjust/ManualAdjust.js
```diff
@@ -83,28 +68,46 @@ class ManualAdjust extends Component {
...
+
   showErrorToast = error => this.props.toastrError(error);
 
   render() {
-    if (!this.props.allowance) return null;
+    const { allowance } = this.props;
+    const allowanceInDays = { ...allowance };
+
+    this.allowanceInDays = this.showAllowanceInDays(allowanceInDays);
```
iscekic: Don't assign `allowanceInDays` to a property.

src/main/frontend/src/store/sagas/leaveProfileSaga.js
```diff
@@ -64,7 +81,16 @@ const createLeaveProfileSaga = function*({ payload }) {
 };
 
 const updateLeaveProfileSaga = function*({ payload }) {
-  yield call(updateLeaveProfileApi(payload));
+  const payloadInDays = { ...payload };
+
+  payloadInDays.entitlement = (payloadInDays.entitlement * 8).toString();
```
iscekic: Why do we need `toString` here?

src/main/frontend/src/pages/administration/components/manual-adjust/ManualAdjust.js
```diff
@@ -83,28 +68,46 @@ class ManualAdjust extends Component {
     });
   };
 
-  closeDrawer = () => this.props.closeDrawer();
+  stringToNumber = obj => {
```
iscekic: Bad naming.

src/main/frontend/src/store/sagas/leaveProfileSaga.js
iscekic: Make an issue for the BE guys to fix this and leave a `TODO` here. 😄

Issue #972, opened 2019-09-30
Trying to quickly get from employee edit screen to dashboard causes white screen


rgerstmajer-ec:
Steps to reproduce:

- Log in as HR Manager
- Go to Administration page
- Select Employees tab
- Select random employee
- Wait for Edit sidebar to load
- Rapidly click on dashboard to get to it as fast as possible

Expected result: Sidebar closes and dashboard is loaded
Actual result: Sidebar closes and white screen appears. Back button does nothing, only refreshing the page shows result.

![Clicking_dashboard_from_employee_edit](https://user-images.githubusercontent.com/54444325/65878617-cc106b00-e38e-11e9-8b92-bf1d127d59c0.gif)

jglugonja:
This issue is resolved with new design.

PR #1033, opened 2019-10-29
f-fix/gh-1017-check-personalData

### LINKED ISSUE #1017: Manually changing leave profile for employee crashes the application
Steps to reproduce (must have HR_MANAGER role):
- Go to administration and select Employees tab
- Find your profile information and change leave profile
- Save the change, and go to the dashboard

Expected result:

- New leave profile is set and the corresponding allowance values are set

Actual result:

- Application crashes with an error message (image below)
Screenshot from 2019-10-23 13-35-11

petarkovac:
Resolves #1017


src/main/frontend/src/components/calendar/calendarUtils.js
```diff
@@ -63,7 +63,7 @@ const checkIfPublicHoliday = (day, publicHolidays, personalData) => {
 const checkIfUserDay = dayObject => {
   const { day, publicHoliday, personalData } = dayObject;
 
-  if (!personalData.length) {
+  if (personalData !== undefined && !personalData.length) {
```
iscekic:
```suggestion
  if (personalData != null && !personalData.length) {
```

This checks for both `null` and `undefined`. 💃

petarkovac:
Resolves:  #972

PR #1041, opened 2019-10-31
Improve date comparison

petarkovac:


src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -271,7 +271,7 @@ private void validateUserIsNotMemberAndApproverOfTeam(User user) {
   }
 
   private void validateDates(LocalDate firstDate, LocalDate secondDate, String errorMessage) {
-    if (firstDate.isBefore(secondDate)) {
+    if (firstDate != secondDate && firstDate.isBefore(secondDate)) {
```
stanislavgrujic: You cannot compare dates by `==`, that only compares reference. You should use `equals` instead.

Issue #1008, opened 2019-10-10
Handle expired carried over hours


petrovicdanilo:
- if an employee doesn't spend his carried over hours until certain date, his allowance should be reduced by unspent amount. That information should be kept in the database and displayed to end-users in the administration, users allowance manual adjust panel.

- the expiration date will be determined subsequently

PR #1044, opened 2019-10-31
b-enhancement/gh-1008-handle-expired-carried-over-days

petrovicdanilo:
Carried over hours scheduled job

- Add scheduled task whose responsibility is to check every evening if carried over hours expiration mechanism is active. If it is active, it checks if the expiration date falls on the given day. If it falls on the given day, it runs a mechanism that reduces user allowance by the number of hours he has carried over from the previous year and has not spent until the expiration date.

- add carriedOverExpirationDate and carriedOverExpiration boolean to 'Year' database table

- fix allowance import logic

Please do not merge after approval, there are frontend changes that need to be implemented in order for this feature to work.


src/main/java/eu/execom/hawaii/jobs/CarriedOverHoursExpirationJob.java
```diff
@@ -0,0 +1,86 @@
...
+    LocalDate expirationDate = year.getCarriedOverHoursExpirationDate();
+    List<Day> userAbsencesDays = getUsersSpentDays(allowance, startDate, expirationDate);
+
+    int spentHoursUntilExpirationDate = allowanceService.calculateHours(userAbsencesDays);
+    int usersCarriedOverHours = allowance.getCarriedOver();
+
+    if (spentHoursUntilExpirationDate < usersCarriedOverHours) {
+      int expiredCarriedOverHours = usersCarriedOverHours - spentHoursUntilExpirationDate;
+      allowance.setExpiredCarriedOver(expiredCarriedOverHours);
+      allowanceRepository.save(allowance);
```
stanislavgrujic: Rather use `allowanceService.save(allowance)`. That way you will use all checks that are inside the service, and not access the repository directly.

src/main/java/eu/execom/hawaii/jobs/CarriedOverHoursExpirationJob.java
```diff
@@ -0,0 +1,86 @@
...
+  private YearRepository yearRepository;
+  private AllowanceRepository allowanceRepository;
+  private DayService dayService;
+  private AllowanceService allowanceService;
+
+  @Autowired
+  public CarriedOverHoursExpirationJob(YearRepository yearRepository, AllowanceRepository allowanceRepository,
+      DayService dayService, AllowanceService allowanceService) {
+    this.yearRepository = yearRepository;
+    this.allowanceRepository = allowanceRepository;
```
stanislavgrujic: In general, it is a good practice to use service references in components, and leave the repository access only to services. That way you will re-use all the logic that is built into the services, for validation of an entity being saved. Also, you will have a clear layer separation.

Issue #1063, opened 2019-11-19
Validate Leave profile input numbers for allowance


petarkovac:
When editing, or creating, a new leave profile doesn't have any validation (example; we can input 10000 days for annual leave).

Suggestion: validate input fields to be less than 365 days.

PR #1065, opened 2019-11-19
b-fix/gh-1063-validate-leave-profile-value

petarkovac:
Resolve: #1063


src/main/java/eu/execom/hawaii/dto/LeaveProfileDto.java
```diff
@@ -23,26 +24,32 @@
...
 
   @NotNull
   @ValidAllowance(message = "'Training' imputed value must be divisible by 4.")
+  @Max(value = 2080, message = "Max training days cannot be more than there are days in a year")
   private int training;
 
   @NotNull
   @ValidAllowance(message = "'Max allowance from next year' imputed value must be divisible by 4.")
+  @Max(value = 2080, message = "Max allowance from next year cannot be more than there are days in a year")
   private int maxAllowanceFromNextYear;
```
stanislavgrujic: Messages are not consistent with messages for @ValidAllowance - in those messages there is a `'` sign around the value being validated, and here it is missing. After you update the messages, you can merge.

PR #1165, opened 2020-01-16
f-enhancement/gh-1164-remove-administration-tab-on-mobile-screens

jprotic:
Resolves: #1164


src/main/frontend/src/components/navigation/Sidebar.js
```diff
@@ -42,7 +42,10 @@ class Sidebar extends Component {
         to={navLink.url}
         onClick={this.closeDrawer}
       >
-        <LinkWrapper className="align-items-center d-flex p-3 p-lg-2">
+        <LinkWrapper
+          isAdministrationLink={navLink.name}
```
fkis1: I think it would make more sense for `isAdministrationLink` prop value be of type `Boolean`. Or to rename the prop so it reflects it's value (something like `navLinkName`). What do you think?

jolpazy: yes, i wanted to say that (:

Issue #1150, opened 2020-01-15
Utility classes should not be instantiated


stefankaticec:
Utility classes should be final and with a private constructor, so that they can't be instantiated. For example: CookeUtils.

PR #1167, opened 2020-01-17
b-enhancement/gh-1150-utility-classes-redefining

nefeyel:
Resolves: #1150


src/main/java/eu/execom/hawaii/util/CookieUtils.java
```diff
@@ -10,7 +10,10 @@
 import java.util.Optional;
 import java.util.stream.Stream;
 
-public class CookieUtils {
+public final class CookieUtils {
+
+  public CookieUtils() {
```
stefankaticec: Constructor needs to be private.

Issue #1162, opened 2020-01-16
Add edit year expired days inputs disabled prop


fkis1:
When editing a year, expired days inputs should be disabled either:
1. If expired days switch wasn't turned on before end of year (case where unused days will be deleted) => disable both expired days switch and date picker after start of new year
2. If expired days switch was turned on before end of year => disable both expired days switch and date picker after the expiration date of expired days.

PR #1168, opened 2020-01-17
f-enhancement/gh-1162-add-edit-year-expired-days-inputs-disabled-prop

fkis1:
Resolves: #1162


src/main/frontend/src/pages/administration/components/years/EditYear.js
```diff
@@ -88,6 +88,41 @@ class EditYear extends Component {
     setSubmitting(false);
   };
 
+  getMostRecentDate = datesArr => new Date(Math.max(...datesArr));
+
+  getCarriedOverHoursExpirationDateMinDate = currentYear => {
+    const firstDayOfSubsequentYearDate = new Date(currentYear + 1, 0, 1);
```
jolpazy: I recommend having a separate js file, or we already have one for date utilities, where we could have all date manipulating functions.

you could have functions like addDaysToDate(number), getMostRecentDate(date, date) isDateBeforeDate etc...

Issue #1020, opened 2019-10-24
Enable HR manager role to be able to add sick leave for employees


petarkovac:
When HR_MANAGER's role tries to add sick leave there should be an option to choose if the sick leave is for themselves or other employees.

PR #1181, opened 2020-01-22
f-enhancement/gh-1020-enable-manager-to-make-requests-for-others

fkis1:
Resolves: #1020, #1182


src/main/frontend/src/components/modal/RequestModal.js
```diff
@@ -139,7 +141,7 @@ class RequestModal extends Component {
         <ModalBody className="py-4 container-fluid overflow-visible">
           {isSpecificRequest && (
             <RequestForm
-              user={user}
+              user={searchedEmployee ? searchedEmployee : user}
```
jolpazy: will `searchedEmployee || user` work

fkis1: `searchedEmployee` can either be an `object` or `null` so it should work. I have changed the ternary operator to logical OR

Issue #1146, opened 2020-01-09
Change Error message when editing Leave type name to existing one


jglugonja:
Steps to reproduce: 
1. Log in with HR credentials
2. Click on Administration many
3. Click on Leave Types tab
4. Click on existing Leave type 
5. Change name of Leave type to already existing
6. Click on button Save

Actual result:

![Screenshot_2020-01-09 Hawaii execom](https://user-images.githubusercontent.com/58041536/72062220-a3760680-32d7-11ea-9b0c-9062a45b9530.png)

Sugestion:
Leave Type with same name already exist.

nefeyel:
This happens also when creating new leave type (Absence) with existing name, not only with editing

jglugonja:
This issue is still reproducible. On the other hand, correct error message is shown when creating new leave type (Absence) with existing name (Absence type: Bonus Days already exists).

jglugonja:
This issue is fixed.

PR #1192, opened 2020-01-23
b-enhancement/gh-1146-change-error-msg-for-existing-leave-type

nefeyel:
resolves: #1146


src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -40,8 +39,13 @@ public AbsenceService(AbsenceRepository absenceRepository) {
    * @return Absence with provided id if exists.
    * @throws EntityNotFoundException if an Absence with given id is not found.
    */
-  public Absence getById(Long id) {
-    return absenceRepository.getOne(id);
+  public Absence findById(Long id) {
```
stefankaticec: Try to follow the convention:
find*** as method name returns an Optional.
get*** returns the value or throws exception.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -51,22 +55,55 @@ public Absence getById(Long id) {
...
+  private void validateAbsenceIsTrainingAndUndeductible(Absence absence) {
     if (absence.isTraining() && !absence.isDeducted()) {
       throw new ActionNotAllowedException("Training and education leave type must be deducted!");
     }
+  }
 
-    return absenceRepository.save(absence);
+  private void validateAbsenceNameIsUnique(Absence absence) {
+    var databaseAbsence = findById(absence.getId());
+    var databaseName = databaseAbsence.getName();
```
stefankaticec: Just name.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -51,22 +55,55 @@ public Absence getById(Long id) {
...
+   * @param absence the Absence entity to be updated.
+   */
   public Absence update(Absence absence) {
+    validateAbsenceIsTrainingAndUndeductible(absence);
+    validateAbsenceNameIsUnique(absence);
+
+    return absenceRepository.save(absence);
+  }
+
+  private void validateAbsenceIsTrainingAndUndeductible(Absence absence) {
```
stefankaticec: Is this method a validation method or a guard?

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -51,22 +55,55 @@ public Absence getById(Long id) {
...
 
-    return absenceRepository.save(absence);
+  private void validateAbsenceNameIsUnique(Absence absence) {
+    var databaseAbsence = findById(absence.getId());
+    var databaseName = databaseAbsence.getName();
+    var updatedAbsenceName = absence.getName();
+    boolean isAbsenceNameChanged = !databaseName.equals(updatedAbsenceName);
+
+    if (isAbsenceNameChanged && absenceRepository.existsByName(absence.getName())) {
+      throw new DataIntegrityViolationException(
```
stefankaticec: Shouldn't data integrity be handled on the database level?

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -51,22 +55,55 @@ public Absence getById(Long id) {
...
+  private void validateAbsenceExistsByName(Absence absence) {
     if (absenceRepository.existsByName(absence.getName())) {
-      throw new DataIntegrityViolationException("Absence type: " + absence.getName() + " already exists.");
+      throw new DataIntegrityViolationException(
+          "Absence type: " + absence.getName() + " already exists.");
     }
+  }
 
-    return absenceRepository.save(absence);
+  private void validateAbsenceIsBonusOrTraining(Absence absence) {
```
stefankaticec: Method name is misleading.

src/test/java/eu/execom/hawaii/service/AbsenceServiceTest.java
```diff
@@ -26,9 +26,7 @@
 import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.BDDMockito.given;
-import static org.mockito.Mockito.doThrow;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.*;
```
stefankaticec: When you have less them 10 imports from the same package, import them individually.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
nefeyel: should I call it "checkAbsence...." istead ?

src/main/java/eu/execom/hawaii/service/AbsenceService.java
nefeyel: https://github.com/execom-eu/hawaii/pull/1192/files#r371192166 same question ?

stefankaticec: Name is misleading because it throws if absence Is bonus (as in the name of the method).

src/main/java/eu/execom/hawaii/service/AbsenceService.java
nefeyel: Done

src/main/java/eu/execom/hawaii/service/AbsenceService.java
stefankaticec: Method name should directly imply what it is doing. We should categorically refuse to go into the method and see it's implementation to understand what the name of the method does. If that should not be possible, we write java doc. 

In this case, we consider what we want to achieve: In order for this request to be fulfilled, we require the absence not to be 'training' and be deductible.
So, this becomes a requireAbsenceIsNotTraining() && requireAbsenceIsDeductable().
By making each function accept a parameter of type Absence, these functions now become Predicates. Predicates have methods that easily form an expression tree (and, or).

Should this project go overboard, we will use these predicates to build a rule engine for validating the requests.

src/test/java/eu/execom/hawaii/service/AbsenceServiceTest.java
nefeyel: then ActionNotAllowedException ?

src/main/java/eu/execom/hawaii/service/AbsenceService.java
nefeyel: Should I use the ActionNotAllowedException ?

stefankaticec: The question isn't what exception should be thrown here, rather why are we throwing it in the first place. I'm fairly sure an unique index will handle this, for all absences.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
nefeyel: private void validateAbsenceIsTrainingAndUndeductible(Absence absence) {
    Predicate<Boolean> reqNotTraining = p -> requireAbsenceIsNotTraining(absence);
    Predicate<Boolean> reqDeductable = p -> requireAbsenceIsDeductable(absence);

    if (reqNotTraining.and(reqDeductable).test(true)) {
      throw new ActionNotAllowedException("Training and education leave type must be deducted!");
    }

private boolean requireAbsenceIsNotTraining(Absence absence) {
    return absence.isTraining();
  }

  private boolean requireAbsenceIsDeductable(Absence absence) {
    return !absence.isDeducted();
  }
  }

is this ok?

stefankaticec: It's more complex than it need be. Drop the predicates for now, but do split these into two methods.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -8,8 +8,8 @@
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.stereotype.Service;
-
 import javax.persistence.EntityNotFoundException;
+import java.util.function.*;
```
stanislavgrujic: Please configure your IDE not to group imports. Put a number of 25 imports, before it groups them.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -41,7 +41,12 @@ public AbsenceService(AbsenceRepository absenceRepository) {
    * @throws EntityNotFoundException if an Absence with given id is not found.
    */
   public Absence getById(Long id) {
-    return absenceRepository.getOne(id);
+    return absenceRepository
+        .findById(id)
+        .orElseThrow(
+            () ->
+                new EntityNotFoundException(
+                    "Absence with id '" + id + "' doesn't exist in database."));
```
stanislavgrujic: How is this different than the original `getOne()` which also throws an `EntityNotFoundException` if it does not find an entity with the given id?

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -51,22 +56,63 @@ public Absence getById(Long id) {
...
-    if (absence.isBonusDays() || absence.isTraining()) {
-      throw new ActionNotAllowedException("It is not allowed to create new bonus or training absences.");
-    }
+    requireAbsenceIsNotBonusNeitherTraining(absence);
+    requireAbsenceExistsByName(absence);
+
+    return absenceRepository.save(absence);
+  }
+
+  private void requireAbsenceExistsByName(Absence absence) {
```
stanislavgrujic: Name of the method is misleading, since it does exactly the opposite - it throws an exception if an entity exists. Please rename to `requireAbsenceNameDoesNotExist`, or something similar.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -51,22 +56,63 @@ public Absence getById(Long id) {
...
     if (absenceRepository.existsByName(absence.getName())) {
       throw new DataIntegrityViolationException("Absence type: " + absence.getName() + " already exists.");
     }
+  }
 
-    return absenceRepository.save(absence);
+  private void requireAbsenceIsNotBonusNeitherTraining(Absence absence) {
```
stanislavgrujic: Please rename to `requireAbsenceIsNotBonusNorTraining`

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -51,22 +56,63 @@ public Absence getById(Long id) {
...
+    if (absence.isBonusDays() || absence.isTraining()) {
+      throw new ActionNotAllowedException(
+          "It is not allowed to create new bonus or training absences.");
+    }
   }
 
+  /**
+   * Updated the provided Absence to repository. It is not allowed to update absences of training
```
stanislavgrujic: Grammar: *updated* -> *update*.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -51,22 +56,63 @@ public Absence getById(Long id) {
...
       throw new ActionNotAllowedException("Training and education leave type must be deducted!");
     }
+  }
 
-    return absenceRepository.save(absence);
+  private boolean requireAbsenceIsNotTraining(Absence absence) {
+    return absence.isTraining();
+  }
+
+  private boolean requireAbsenceIsDeductible(Absence absence) {
```
stanislavgrujic: These two methods do not help improve code readability. 
If you have a convention that methods with prefix *require_* throw an exception if the required condition is not met, you would have to throw an exception inside these two methods.
Also, it is a good practice to name methods which return boolean value by using verbs `to be` or `to have`, so these methods should be named `isAbsenceNonDeductible()` and `isAbsenceTraining()` (**pay attention to what is negated and to use proper naming!**), which just wraps the similar method calls. Please refactor to:
```
private void requireTrainingIsDeductible(Absence absence) {
  if (absence.isTraining() && !absence.isDeducted()) {
    throw new ActionNotAllowedException("Training and education leave type must be deducted!");
  }
}
```

Always think what your method is doing and what would be the proper naming. In this case, the method verifies that if an absence is training, it must not be deductible, therefore, the proper naming would be `requireTrainingIsDeductible`.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -51,22 +56,63 @@ public Absence getById(Long id) {
...
+    return !absence.isDeducted();
+  }
+
+  private void requireAbsenceNameIsUnique(Absence absence) {
+    var databaseAbsence = getById(absence.getId());
+    var name = databaseAbsence.getName();
+    var updatedAbsenceName = absence.getName();
+    boolean isAbsenceNameChanged = !name.equals(updatedAbsenceName);
```
stanislavgrujic: Instead of adding this piece of code and a TODO, add a new database change set which adds unique constraint on the `name` column in the 'absence' table.

nefeyel: In table Absence, name is already unique, so i don't need this peace of code?

stanislavgrujic: Maybe you can test it by using the application?

nefeyel: works fine without it...

stanislavgrujic: Please remove it from the code base, in that case.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -56,53 +50,48 @@ public Absence getById(Long id) {
...
-    requireAbsenceIsNotBonusNeitherTraining(absence);
-    requireAbsenceExistsByName(absence);
+    requireAbsenceIsNotBonusNorTraining(absence);
+    requireAbsenceNameDoesNotExist(absence);
 
     return absenceRepository.save(absence);
   }
 
-  private void requireAbsenceExistsByName(Absence absence) {
+  private void requireAbsenceNameDoesNotExist(Absence absence) {
```
stanislavgrujic: If the uniqueness constraint is added to the table in the database, this check is not needed. Please remove and check if the application behaves the same way.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -56,53 +50,48 @@ public Absence getById(Long id) {
...
+   * Update the provided Absence to repository. It is not allowed to update absences of training
    * and undeductible types. Also, to update absence type's name, absence name needs to be unique.
    *
    * @param absence the Absence entity to be updated.
    */
   public Absence update(Absence absence) {
-    requireAbsenceIsNotTrainingAndDeductible(absence);
-    requireAbsenceNameIsUnique(absence);
+    requireTrainingIsDeductible(absence);
+//    requireAbsenceNameIsUnique(absence);
```
stanislavgrujic: Please do not leave commented out code in the code base.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -56,53 +50,48 @@ public Absence getById(Long id) {
...
 
-    return absenceRepository.save(absence);
+    try{
+      return absenceRepository.save(absence);
+    }catch(DataIntegrityViolationException e) {
```
stanislavgrujic: Please apply proper formatting. Is there a need to catch and re-throw this excption?

src/main/java/eu/execom/hawaii/service/AbsenceService.java
```diff
@@ -114,7 +103,7 @@ private void requireAbsenceNameIsUnique(Absence absence) {
       throw new IllegalStateException("Absence type: " + absence.getName() + " already exists.");
     }
   }
-
+*/
```
stanislavgrujic: It is a bad practice to leave commented out code. Please delete it.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
nefeyel: You're right. It is unneeded. Application behaves the same way.

src/main/java/eu/execom/hawaii/service/AbsenceService.java
nefeyel: Well yes because my final goal is to override existing exception message [(https://user-images.githubusercontent.com/58041536/72062220-a3760680-32d7-11ea-9b0c-9062a45b9530.png)]

stanislavgrujic: Then just apply the proper formatting.

nefeyel: applied execom formater on whole IDE

src/test/java/eu/execom/hawaii/service/AbsenceServiceTest.java
```diff
@@ -74,14 +74,14 @@ public void shouldGetAllAbsences() {
   public void shouldGetAbsenceById() {
     // given
     var absenceId = 1L;
-    given(absenceRepository.getOne(absenceId)).willReturn(mockAbsence);
+    given(absenceRepository.findById(absenceId)).willReturn(Optional.of(mockAbsence));
```
stanislavgrujic: Is this change still valid?

nefeyel: reverted, works fine

Issue #1153, opened 2020-01-15
Overuse of streams


stefankaticec:
Some parts of code overuse streams, leading to code that is inefficient. 
For example: 
stream().filter().filter()
which effectively means that stream will iterate twice through the collection. Considering that filters use [Predicates](https://docs.oracle.com/javase/8/docs/api/java/util/function/Predicate.html) and other stream methdos use [Functions](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html), we can use **and**/**or** methods on predicates and **andThen** on Functions to combine them into a single function.

PR #1222, opened 2020-02-26
b-enhancement/gh-1153-overuse-of-streams

slobodanc93:
Resolves: #1153


src/main/java/eu/execom/hawaii/batch/AppogeeLeaveProcessor.java
```diff
@@ -177,9 +177,9 @@ private Request findRequestForUpdate(AppogeeLeaveRequestImportData leaveImport,
     Optional<Absence> absence = getAbsence(leaveImport.getType());
 
     Stream<Request> requestStream = allUsersRequests.stream()
-                                                    .filter(requestsSubmittedOnDate(dateSubmitted))
-                                                    .filter(request -> isBetween(firstDay, lastDay, request))
-                                                    .filter(requestsWithDifferentDuration(leaveImport));
+                                                    .filter(requestsSubmittedOnDate(dateSubmitted)
+                                                            .and(request -> isBetween(firstDay, lastDay, request))
+                                                            .and(requestsWithDifferentDuration(leaveImport)));
```
stanislavgrujic: These are the same level filter values, so they should not be indented. Please fix indentation in all changed files.

Issue #1149, opened 2020-01-15
Conversion from JSON leads to logical issues


stefankaticec:
AuditInformationService:deserialize method creates an empty instance (via reflection) if it fails to convert to json for given class. This brings no value and can create issues down the road when interpreting data from an audit log. This piece of code has to go. 

The disappearance of the a class from the code base is not handled at all, so a safe alternative would be to convert the JSON to Map<String, Object> when data can't be parsed to Class<T>.

PR #1224, opened 2020-02-28
b-fix/gh-1149-conversion-from-JSON-leads-to-logical-issues

slobodanc93:
Resolves: #1149 

Situation 1: Certain fields are not recognized (e.g. "job_title":"Developer" instead of "jobTitle":"Developer"). This kind of error will be logged.
Situation 2: Whole entity is not recognized (e.g. "auditedEntity":"USER_X" instead of "auditedEntity":"USER"). This kind of error will be logged, but also custom exception will be thrown (which means that end-user will see an error message on fronted to contact support).


src/main/java/eu/execom/hawaii/service/AuditInformationService.java
```diff
@@ -156,9 +158,16 @@ private boolean isUserArchivedOrRestored(AuditInformation auditInformation) {
   private <T extends Audit> T deserialize(String currentValue, Class<T> clazz) {
     try {
       return OBJECT_MAPPER.readValue(currentValue, clazz);
+    } catch (UnrecognizedPropertyException e) {
+      log.error("Unrecognized property in JSON representation: {}", e.getMessage());
+      OBJECT_MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
stefankaticec: Since this is a single object, it makes no sense to configure it over and over again. Use a static block to configure a static object.

src/main/java/eu/execom/hawaii/service/AuditInformationService.java
```diff
@@ -156,9 +158,16 @@ private boolean isUserArchivedOrRestored(AuditInformation auditInformation) {
...
+      return OBJECT_MAPPER.readValue(currentValue, clazz);
     } catch (IOException e) {
-      log.error("Failed to map JSON representation of entity '{}' as object", clazz.getName());
-      return clazz.getDeclaredConstructor().newInstance();
+      Map unrecognizedObject = OBJECT_MAPPER.readValue(currentValue, Map.class);
+      log.error("Failed to map JSON representation to object: {}", unrecognizedObject);
+      throw new AuditDataParseException("Audit logs are not loaded correctly, please contact IT support.");
+    } finally {
+      OBJECT_MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, true);
```
stefankaticec: you wont need this.

Issue #1216, opened 2020-02-12
Inactive Year created in Postman could not be deleted


jglugonja:
Steps to reproduce:
1. Send POST request in Postman (https://hawaii2.execom.eu/api/years) with following JSON in the body of request:

{
  "active": false,
  "carriedOverHoursExpirable": true,
  "carriedOverHoursExpirationDate": "2024-04-01",
  "id": 18,
  "year": 2023
}

2. Try to delete created Year.

Actual result: 
Year could not be deleted with popup error message
![Screenshot_2020-01-09 Hawaii execom](https://user-images.githubusercontent.com/58041536/74332199-854b6c80-4d95-11ea-8a2c-3ac36023665d.png)

 

P.S. You will create Year that could be deleted as expected if you send: 

{
    "id": "18",
    "year": "2023",
    "active": "false",
    "carriedOverHoursExpirable": "false",
    "carriedOverHoursExpirationDate": null
}

slobodanc93:
These two API calls will not make an issue (order of fields is not relevant) and they will add new resources with the latest available ID.
But...POST requests can also update existing resources if we provide valid and existing ID.
If we send a request to add the year 2050 with "id": 1, it will update any value that was in a row with id 1. So later, we won't be able to delete the year 2050 because it has been used in the allowance table.
As a solution, if necessary, we could forbid updates via POST (and update only via PUT), but we would have to check it with the frontend team because they might be using POST for updating.

PR #1233, opened 2020-03-06
b-fix/gh-1216-Inactive-Year-created-in-Postman-could-not-be-deleted

slobodanc93:
Resolves: #1216 
Frontend validation when editing the year is already in place. With these changes, APIs are also protected from making unwanted changes.


src/main/java/eu/execom/hawaii/service/YearService.java
```diff
@@ -96,6 +100,16 @@ private void validateCarriedOverExpirationActivation(Year year) {
     }
   }
 
+  private void validateNameStatus(Year year) {
```
stanislavgrujic: I would rather name this method `validateYearChange`, since if we add more validations, the name could become verbose or obsolete.

slobodanc93:
@stanislavgrujic I just noticed that tests should also be slightly updated according to new validation when saving a year. Please have a look at this before I merge, thank you.

Issue #1208, opened 2020-01-27
Log in issue after dropping the database


jglugonja:
Steps for reproducing the issue:
1. User is logged in with his credentials on his mobile phone. 
2. Database is dropped. 
3. User goes again to application URL from his mobile phone.

Actual result:
User is logged in with another users credentials.
P.S. Another user has never been logged in on particular mobile device.

![screenshot-1580127845267](https://user-images.githubusercontent.com/58041536/73181260-2e982e80-4117-11ea-84a2-82ae69681f79.jpg)

PR #1236, opened 2020-03-10
b-fix/gh-1208-log-in-issue-after-dropping-the-database

popovicbojan:
Resolves: #1208


src/main/java/eu/execom/hawaii/security/CustomUserDetailsService.java
```diff
@@ -36,4 +36,11 @@ public User loadUserById(Long id) {
                          .orElseThrow(
                              () -> new EntityNotFoundException("User with given id doesn't exist in database."));
   }
+
+  @Transactional
```
stanislavgrujic: Why do you need this to be transactional? There is only a read command over the database.

src/main/java/eu/execom/hawaii/security/UserPrincipal.java
```diff
@@ -37,6 +37,8 @@ Long getId() {
     return id;
   }
 
+  String getEmail() { return email; }
```
stanislavgrujic: Please do not inline method body. Always try to keep the code style that is used throughout the codebase which you joined.

src/main/java/eu/execom/hawaii/security/CustomUserDetailsService.java
popovicbojan: I saw that methods loadUserByUsername() and loadUserById() above are all transactional, so I thought that all similar methods needed to be transactional, although it doesn't make a sense. Should I remove transactional annotation from all of them?

src/main/java/eu/execom/hawaii/security/UserPrincipal.java
popovicbojan: Ok, I'll correct this.

src/main/java/eu/execom/hawaii/security/CustomUserDetailsService.java
stanislavgrujic: Please do. :)

Issue #1215, opened 2020-02-06
Add Approver's leaves to team calendar 


jglugonja:
End users (employees) suggested new enhancement. 
User should be able to see Team leader`s leaves in Team calendar.

jglugonja:
Team leader's leaves are not visible in List view.
![NotClickable](https://user-images.githubusercontent.com/58041536/76319168-1a7f4980-62df-11ea-9f8f-29deb0b725cf.gif)

popovicbojan:
This commit also fixes #1242

PR #1240, opened 2020-03-10
b-enhancement/gh-1215-add-approvers-leaves-to-team-calendar

popovicbojan:
Resolves: #1215


src/main/java/eu/execom/hawaii/api/controller/RequestController.java
```diff
@@ -143,7 +144,10 @@ public RequestController(RequestService requestService, UserService userService)
       @RequestParam(required = false) Long teamId,
       @RequestParam(required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) YearMonth yearMonth) {
     //@formatter:on
-    var requests = requestService.findAllByTeamByMonthOfYear(teamId, yearMonth);
+    var teamMembersRequests = requestService.findAllByTeamByMonthOfYear(teamId, yearMonth);
+    var teamApproversRequests = requestService.findAllByTeamApproversWithinDates(teamId, yearMonth);
+    var requests = new ArrayList<>(teamMembersRequests);
+    requests.addAll(teamApproversRequests);
```
stanislavgrujic: There is no need to explicitly create a new ArrayList from teamMembersRequests. Please just add teamApproversRequests:
```
var teamApproversRequests = requestService.findAllByTeamApproversWithinDates(teamId, yearMonth);
requests.addAll(teamApproversRequests);
```

popovicbojan: Ok, I'll correct it.

PR #1243, opened 2020-03-10
 b-enhancement/gh-1215-add-approvers-leaves-to-team-calendar

popovicbojan:
Resolves: #1215


src/main/java/eu/execom/hawaii/api/controller/RequestController.java
```diff
@@ -145,8 +145,11 @@ public RequestController(RequestService requestService, UserService userService)
       @RequestParam(required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) YearMonth yearMonth) {
     //@formatter:on
     var requests = requestService.findAllByTeamByMonthOfYear(teamId, yearMonth);
-    var teamApproversRequests = requestService.findAllByTeamApproversWithinDates(teamId, yearMonth);
-    requests.addAll(teamApproversRequests);
+    if (teamId != null) {
```
stanislavgrujic: When can this be null? It does not make sense to request team's leave requests without providing the teamId.

popovicbojan: Switching to team calendar and requesting all employees sends request without teamId. 
Request URL:http://localhost:8080/api/requests/team/month?yearMonth=2020-03&teamId=

stanislavgrujic: That is semantically wrong. :) OK, leave it as is, then.

Issue #1120, opened 2019-12-06
When trying to edit inactive User error pops up


jglugonja:
1.Click on Administration
2.Go to Employees tab
3.Click on Inactive users icon
4.Click on edit user 

Edit tab opens and Error pops up

fkis1:
Unable to reproduce the issue. Can you please provide additional info?

jglugonja:
1.Create sheduled user.
2.Leave it inactive.
3.Go to Employees tab
4.Click on Inactive users 
4.Click on edit user

After clicking on inactive(sheduled user) Error message pops up.

fkis1:
This issue occurs due to newly created scheduled employees not having allowance. Allowance is added once an employee becomes active. Without allowance, the edit drawer for an employee cannot be populated with data, hence the error.

One solution to this issue is to add allowance to newly created scheduled employees, since inactive employees don't have access to the app, and therefore cannot make requests.

PR #1246, opened 2020-03-11
b-fix/gh-1120-when-trying-to-edit-inactive-user-error-pops-up

popovicbojan:
Resolves: #1120


src/main/java/eu/execom/hawaii/api/controller/AllowanceController.java
```diff
@@ -75,6 +76,12 @@ public AllowanceController(AllowanceService allowanceService, UserService userSe
       @ApiIgnore @AuthenticationPrincipal User authUser, @RequestParam(required = false) Long userId,
       @RequestParam(required = false) Integer year) {
     Long requestedUserId = userId == null ? authUser.getId() : userId;
+
+    User user = userService.findById(requestedUserId);
+    if(!user.isActive()){
```
stanislavgrujic: Please configure your IDE to format files properly. This is not good formatting.

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -68,6 +69,13 @@
     return errorResponse(ex.getMessage(), HttpStatus.NOT_FOUND);
   }
 
+  @ExceptionHandler({NoContentException.class})
+  public ResponseEntity<Map<String, String>> handle(HttpServletRequest request, NoContentException ex){
```
stanislavgrujic: Please reformat the file after you configure your IDE.

src/main/java/eu/execom/hawaii/api/controller/AllowanceController.java
```diff
@@ -75,6 +76,12 @@ public AllowanceController(AllowanceService allowanceService, UserService userSe
...
+      throw new NoContentException(user.getFullName() + " is not active user.");
```
stanislavgrujic: The naming of the exception is not correct -> you should indicate what the cause is (user is inactive), not what the end result is (status code of NO_CONTENT). Please rename this exception to `InactiveUserException` which gives more information about the actual cause of the issue.

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -68,6 +69,13 @@
...
+    logException(request, ex);
+
+    return errorResponse(ex.getMessage(), HttpStatus.NO_CONTENT);
```
stanislavgrujic: The NO_CONTENT status code belongs to 2xx status code, which in general mean that the request is properly processed. In this case it is not the case, since the request cannot be properly processed due to the user being inactive (it is a client error requesting an inactive user). I think more appropriate here is to use 4xx status code (e.g., `403 - Forbidden` or `406 - Not Acceptable` or `422 - Uprocessable Entity`. Please investigate and use the one which you think is the most appropriate.

popovicbojan:
Scheduled users don't have allowance yet, so by requesting nonexistent allowance AllowanceNotFoundException will be thrown and `404 - Not Found` status code will be returned.

src/main/java/eu/execom/hawaii/service/AllowanceService.java
```diff
@@ -107,7 +108,11 @@ public void applyPendingRequest(Request request, boolean requestCanceled) {
    */
   public Allowance getByUserAndYear(Long userId, Integer year) {
     int allowanceForYear = year == null ? LocalDate.now().getYear() : year;
-    return allowanceRepository.findByUserIdAndYearYear(userId, allowanceForYear);
+    Allowance allowance = allowanceRepository.findByUserIdAndYearYear(userId, allowanceForYear);
+    if (allowance == null) {
+      throw new AllowanceNotFoundException("Allowance for user with id " + userId + " not found.");
```
stanislavgrujic: Please wrap the userId in single quotes, so that it is clear that it is a parameter:
`Allowance for user with id 'userId' not found.`

src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -68,6 +69,13 @@
     return errorResponse(ex.getMessage(), HttpStatus.NOT_FOUND);
   }
 
+  @ExceptionHandler(AllowanceNotFoundException.class)
+  public ResponseEntity<Map<String, String>> handle(HttpServletRequest request, AllowanceNotFoundException ex){
```
stanislavgrujic: This line is not properly formatted: `){` -> `) {`.

Issue #1225, opened 2020-02-28
Dashboard design changes


slobodanc93:
Both public holidays and bonus days are represented with a **star** image, **public holidays** icon needs to be changed (a new icon will be delivered).
**Background color for the request** (yellow/green/red) need to have lower opacity (50%)

PR #1250, opened 2020-03-11
f-enhancement/gh-1225-dashboard-design-changes

MiljkovicVeljko:
Resolves: #1225


src/main/frontend/src/components/calendar/Day.js
```diff
@@ -83,7 +83,7 @@ const ImageContainer = styled.div`
 `;
 
 const Image = styled.img`
-  width: 12px;
+  width: ${props => (props.holiday ? '22px' : '12px')};
```
iscekic: Double check that this won't cause alignment issues.

Issue #1189, opened 2020-01-23
Users without Start professional career date don't have Carried over days


jglugonja:
Imported users without Start professional career date will not have Carried over days. For this users Carried over days will be updated automatically by hrassistant@execom.eu on January 1st (visible in Audit tab) but this change will not be visible. Actual result shows allowance without Carried over days. 

![no_carried_over-days](https://user-images.githubusercontent.com/58041536/72984656-8ae01300-3de4-11ea-96e7-7b1b099a8a58.gif)

slobodanc93:
Currently on hold. Needs to be re-tested after the closing of #1213.

slobodanc93:
File **allowance-import.csv** if present will override existing values in the database.
So basically the JOB scheduled for 1st January was successful and updated carried over days, but data in the import file later changed carried over days to initial values.
**The solution** is to implement a validation when importing allowances to check if data is already in the database (based on a combination of user and year).

PR #1252, opened 2020-03-12
 b-fix/gh-1189-Users-without-Start-professional-career-date-don't-have-Carried-over-days

slobodanc93:
Resolves: #1189


src/main/java/eu/execom/hawaii/batch/AllowanceDbWriter.java
```diff
@@ -20,7 +23,16 @@ public AllowanceDbWriter(AllowanceRepository allowanceRepository) {
 
   @Override
   public void write(List<? extends Allowance> importedAllowances) throws Exception {
-    allowanceRepository.saveAll(importedAllowances);
+    var currentAllowances = new ArrayList<>(allowanceRepository.findAll());
+    var newAllowances =
+        importedAllowances.stream()
```
stanislavgrujic: Please leave this line as in the same level as the left side of the assignment operator.

src/main/java/eu/execom/hawaii/batch/AllowanceDbWriter.java
```diff
@@ -20,7 +23,16 @@ public AllowanceDbWriter(AllowanceRepository allowanceRepository) {
...
+            .filter(isNewAllowance(currentAllowances))
+            .collect(Collectors.toList());
+    newAllowances.forEach(allowance -> log.debug("Data saved for allowance '{}'.", allowance));
```
stanislavgrujic: Is this logging really needed? My assumption is that it will never be read, and yet we will iterate through possibly a lot of allowance records just in order to have it written to a log file.

slobodanc93: I agree. Logging is unnecessary here so I removed it.

Issue #1262, opened 2020-03-16
Validation for calculations


jglugonja:
Include validation in calculations to prevent negative numbers

![Screenshot from 2020-03-16 14-58-11](https://user-images.githubusercontent.com/58041536/76767029-3a0fe980-6799-11ea-9087-572b4e877381.jpg)

PR #1263, opened 2020-03-17
 b-fix/gh-1262-validation-for-calculations

popovicbojan:
Resolves: #1262


src/main/java/eu/execom/hawaii/exceptions/NegativeAllowanceException.java
```diff
@@ -0,0 +1,8 @@
+package eu.execom.hawaii.exceptions;
+
+public class NegativeAllowanceException extends RuntimeException {
+
+  public NegativeAllowanceException(String message) {
```
stanislavgrujic: Instead of passing the whole message, you could do the following:
```
private static final String MESSAGE = "Allowance for user '%s' can be manually adjusted for a minimum of '-%f.1' days.";

public NegativeAllowanceException(User user, int remainingAllowance) {
  super(String.format(MESSAGE, user.getFullName, remainingAllowance / 8);
}

```
Also, value of remainingAllowance /8 needs to be rounded to either 0.5 or 0, since we do not allow modification of less than half a day.

Issue #1212, opened 2020-01-29
Authentication token is incorrect


stefankaticec:
Token validation is incorrect: it does not validate that the token did not expire. 
Once issued, token is valid forever, acting like a magic key for a user!
This bug most likely introduced another bug in the client application: the token is not refreshed, ever. 

Contents of the authentication token is incorrect.
Missing information:
1. Issuer (hawaii)
2. Subject (identity of the principal)

PR #1266, opened 2020-03-17
b-fix/gh-1212-Authentication-token-is-incorrect

slobodanc93:
Resolves: #1212


src/main/java/eu/execom/hawaii/security/TokenProvider.java
```diff
@@ -30,24 +35,40 @@ public String createToken(Authentication authentication) {
...
     return Jwts.builder()
                .setSubject(email)
+               .setIssuer(tokenIssuer)
                .setIssuedAt(currentTime)
                .setExpiration(expirationTime)
                .signWith(SignatureAlgorithm.HS512, appProperties.getAuth().getTokenSecret())
                .compact();
   }
 
+  private Long getTokenExpirationMillisecondsValue() {
```
stanislavgrujic: This check is surplus. It is a matter of configuration to set the token expiration properly, and we should not add code which checks configuration. Please remove it together with the created exception class.

Issue #1254, opened 2020-03-13
Security Alert - Buffer Overflow


jglugonja:
Buffer overflow errors are characterized by the overwriting of memory spaces of the background web process, which should have never been modified intentionally or unintentionally. Overwriting values of the IP (Instruction Pointer), BP (Base Pointer) and other registers causes exceptions, segmentation faults, and other process errors to occur. Usually these errors end execution of the application in an unexpected way.

Vulnerable URLs:

1. https://hawaii2.execom.eu/api/users

2. https://hawaii2.execom.eu/api/years?page=0&size=30&sort=year%2Cdesc

PR #1267, opened 2020-03-17
b-fix/gh-1254-Security-Alert-Buffer-Overflow

slobodanc93:
Resolves: #1254


src/main/java/eu/execom/hawaii/exceptions/SortParameterTooLongException.java
```diff
@@ -0,0 +1,10 @@
+package eu.execom.hawaii.exceptions;
+
+public class SortParameterTooLongException extends RuntimeException {
+
+  private static final long serialVersionUID = 3279745025779020740L;
+
+  public SortParameterTooLongException(String message) {
```
stanislavgrujic: You can also internalise this message as:
```
public SortParameterTooLongException(String message) {
  super("Sort parameter is too long.");
}
```

That way you don't have to clutter the invocation place with the message which is not its responsibility anyway.

Issue #1151, opened 2020-01-15
PUT mappings do not conform to rest principles


stefankaticec:
Most HTTP PUT mappings don't have an Identifier in the URL, which isn't conforming to REST principles. 

POST should map to the pool ('/')
GET maps to resource in the pool ('/{id}')
DELETE maps to resource in the pool ('/{id}')
PUT and PATCH should map to resource in the pool ('/{id}')

This makes even more sense when you consider what is in the body of the request: an id. An ID can't be changed by the update request.

PR #1272, opened 2020-03-19
feature/gh-1151-PUT-mappings-do-not-conform-to-rest-principles

slobodanc93:
Resolves: #1151


src/main/java/eu/execom/hawaii/exceptions/PutRequestIdsNotTheSameException.java
```diff
@@ -0,0 +1,9 @@
+package eu.execom.hawaii.exceptions;
+
+public class PutRequestIdsNotTheSameException extends RuntimeException {
```
stanislavgrujic: Please rename to `MismatchingRequestIdsException`. 
As a general rule of thumb for naming:
1. I would avoid using HTTP method in exception name, because it does not give an actionable detail and just adds verboseness to the class name.
2. Do not using articles *The* and *A* in naming.

src/main/java/eu/execom/hawaii/api/controller/AbsenceController.java
```diff
@@ -72,8 +73,12 @@ public AbsenceController(AbsenceService absenceService, IconService iconService)
     return new ResponseEntity<>(new AbsenceDto(absence), HttpStatus.CREATED);
   }
 
-  @PutMapping
-  public ResponseEntity<AbsenceDto> updateAbsence(@RequestBody @Valid AbsenceDto absenceDto) {
+  @PutMapping("/{id}")
+  public ResponseEntity<AbsenceDto> updateAbsence(
+      @PathVariable Long id, @RequestBody @Valid AbsenceDto absenceDto) {
+    if (!id.equals(absenceDto.getId())) {
```
stanislavgrujic: Please investigate introducing a `Filter`/`HandlerInterceptor` class which would check this in a single place, instead of extending all PUT methods with the identical code. You would probably need to introduce a common interface in all DTOs to be able to extract `getId()` method.

src/main/java/eu/execom/hawaii/exceptions/PutRequestIdsNotTheSameException.java
slobodanc93: This exception is removed, as agreed we will ignore ID provided in the request body and update resource based on identifier provided in URL.

Issue #1251, opened 2020-03-12
Add scheduled for deletion icon


jglugonja:
Add icon for users which are scheduled for deletion. Clicking on that icon will open date picker with selected date.
![Scheduled for deletion](https://user-images.githubusercontent.com/58041536/76532032-5862a600-6476-11ea-8a0b-3946f63c97d9.gif)

PR #1277, opened 2020-03-24
enhancement/gh-1251-add-scheduled-for-deletion-icon

popovicbojan:
Resolves: #1251 
Added icon for employees scheduled for deletion. Clicking on icon will open popup displaying information about scheduled action with option of canceling deletion. Canceling deletion will be displayed in employee edit history. Corresponding messages are added.


src/main/java/eu/execom/hawaii/model/audit/Audit.java
```diff
@@ -23,11 +23,15 @@
       field.setAccessible(true);
 
       try {
-        if (field.get(audit) != null && !field.get(audit).equals(field.get(this))) {
+        if ((field.get(audit) != null && !field.get(audit).equals(field.get(this))
+            || ("stoppedWorkingAtExecomDate".equals(field.getName())
+                && field.get(this) != null
+                && field.get(audit) == null))) {
```
stanislavgrujic: This is very complex and hard to understand addition to the condition. Could you extract this part to a variable with a meaningful name?

src/main/java/eu/execom/hawaii/model/audit/Audit.java
```diff
@@ -38,7 +42,8 @@ private AuditedChanges getAuditedChanges(Audit audit, Field field) throws Illega
     AuditedChanges change = new AuditedChanges();
     change.setFieldName(field.getName());
     change.setOldObjectValue(field.get(this) != null ? field.get(this).toString() : "");
-    change.setNewObjectValue(field.get(audit).toString());
+    String newValue = field.get(audit) == null ? "CANCELED" : field.get(audit).toString();
```
stanislavgrujic: Why do we even have to enter specific logic for audit of this change?

src/main/java/eu/execom/hawaii/api/controller/UserController.java
```diff
@@ -242,6 +242,12 @@ public ResponseEntity activateUser(@Valid @ApiIgnore @AuthenticationPrincipal Us
     return new ResponseEntity(HttpStatus.OK);
   }
 
+  @PutMapping("/canceldelete/{id}")
+  public ResponseEntity cancelDelete(@Valid @ApiIgnore @AuthenticationPrincipal User authUser, @PathVariable Long id) {
+    userService.cancelDelete(id, authUser);
+    return new ResponseEntity(HttpStatus.NO_CONTENT);
+  }
```
stanislavgrujic: This action and `activate` are similar by nature. Please update that one to also reflect this approach and use path variable, instead of the request body.

src/main/java/eu/execom/hawaii/model/audit/Audit.java
popovicbojan: When scheduled delete is canceled, new value of `stoppedWorkingAtExecomDate `field is set do null, so instead of calling `null.toString()`, new value will be set to `'CANCELED'`. The same check is done above for `change.setOldObjectValue()` for opposite situation, when old value is null.

src/main/java/eu/execom/hawaii/api/controller/UserController.java
popovicbojan: I think that @slobodanc93 corrected this in his pull request for issue #1272.

src/main/java/eu/execom/hawaii/model/audit/Audit.java
stanislavgrujic: But this will put CANCELED to all new values of null, for any field which allows it, not just this one.
I would suggest to do it like this:
```
String newValue = field.get(audit) == null ? "" : field.get(audit).toString();
```

Issue #1147, opened 2020-01-09
Use different formula for annual allowance calculation


jglugonja:
According to Article 72 of Employment Act (Republic of Serbia) an employee is entitled to one-twelfth of the annual leave under Article 69 of this Act (proportionate part) for each month of work in the calendar year in which he concluded his employment relationship or in which his employment relationships is terminated. 

Implement this formula for annual allowance calculation: 
Annual quota divided by number of months in a Year and multiplied by months left to the end of a Year

PR #1278, opened 2020-03-25
b-enhancement/gh-1147-use-different-formula-for-annual-allowance-calculation

popovicbojan:
Resolves: #1147


src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -432,39 +433,58 @@ private int calculateYearsOfService(User user) {
...
+   * as annual leave.
    */
   private Allowance createAllowance(User user, Year year) {
     LeaveProfile leaveProfile = leaveProfileRepository.getOne(user.getLeaveProfile().getId());
     Allowance allowance = new Allowance();
     allowance.setUser(user);
     allowance.setYear(year);
 
-    if (user.getStartedWorkingAtExecomDate().getYear() != year.getYear()) {
+    @NotNull
```
stanislavgrujic: Are you sure that the effect is the same? Even if it works (and my assumption is that it doesn't, since these annotations are used for bean validation, not for flow control), I am against cluttering the codebase with in method annotations. Please revert, or change the code accordingly.

popovicbojan: This annotation was inserted by IntelliJ when I extracted `user.getStartedWorkingAtExecomDate()` to a variable, because it was repeating in lot of places. I mistake it with Lomboks `@NonNull` so I didn't remove it. Is it ok if I just remove annotation, because `startedWorkingAtExecomDate` filed is mandatory and can't be null?

stanislavgrujic: Sure.

Issue #1276, opened 2020-03-20
Leave profile name validation


jglugonja:
Steps to reproduce:
1. Log in with HR credentials 
2. Click on Administration tab
3. Click on Leave profiles tab
4. Create Leave profile
5. Create another leave profile with the same name

Actual result:

There are two leave profiles with exactly same name

![Screenshot_2020-03-20 Hawaii execom](https://user-images.githubusercontent.com/58041536/77156716-3a152f80-6aa0-11ea-9de2-72fd0ffb7893.png)

PR #1285, opened 2020-03-27
b-fix/gh-1276-leave-profile-name-validation

popovicbojan:
Resolves: #1276


slobodanc93:
Maybe we should add DB control as well. The name field is already annotated to be unique, but it is not working as we're missing constraints in the InitialDbSchema file.

src/main/java/eu/execom/hawaii/service/LeaveProfileService.java
```diff
@@ -10,6 +10,7 @@
 import org.springframework.data.domain.Pageable;
 import org.springframework.stereotype.Service;
 
+import javax.persistence.EntityExistsException;
```
stanislavgrujic: Please remove unused import.

Issue #1306, opened 2020-04-22
Small alignment changes


slobodanc93:
Hawaii logo in the navbar needs to be centered (currently, it is slightly misaligned closer to left).
The new request button is also slightly misaligned, not in line with the content bellow.

PR #1307, opened 2020-04-22
frontend/gh-1306-small-alignment-changes

popovicbojan:
Resolves: #1306


src/main/frontend/src/components/header/styled.js
```diff
@@ -37,6 +37,6 @@ export const Heading = styled.h5`
 
 export const CenteredLogo = styled.div`
   top: 100%;
-  left: 50%;
+  left: 54%;
```
aberonja, line 40: 54%? Why not 50%? This will not be in the middle if it's 54%, right?

popovicbojan: If it is set to 50% then right logo edge will be in the middle, with 54% center of the logo is in the middle.

aberonja: Could you share a screenshot? It sounds odd for me, maybe the logo is not properly cropped?

popovicbojan: The logo is cropped fine, it's because of the right property nature. The right property is used to set the right margin edge of the element and the right edge of its containing block. So it defines the right logo edge position and not the center of the logo.

aberonja3ap: Sounds like something else (right property) is causing that behavior. Please, give it 10 more minutes to figure it out and if this is the only way then do it like this. Thanks :)

I'm giving you the green light, so it's up to you.

popovicbojan: This is the only solution I came up with. :)
I noticed that you approved changes from your other GitHub account and yesterday requested changes from another so merge is still blocked. :)

Issue #1305, opened 2020-04-21
Add allowance overview graph on Dashboard


slobodanc93:
This will be used instead of info cards. Allowance should include leave, training, and sickness.

PR #1308, opened 2020-04-22
f-enhancement/gh-1305-Add-allowance-overview-graph-on-Dashboard

slobodanc93:
Resolves: #1305 
Old and new dashboard


src/main/frontend/src/pages/dashboard/components/styled.js
```diff
@@ -1,14 +1,18 @@
...
   box-shadow: 3px 3px 10px #d3d3d3;
 
   @media (min-width: 767px) {
-    min-height: 390px;
+    min-height: 415px;
   }
 `;
 
+export const DashboardChartContainer = styled.div`
+  min-height: 415px;
```
aberonja, line 13: Shouldn't this be a flex element so it doesn't have to have height?

src/main/frontend/package.json
```diff
@@ -13,6 +15,7 @@
     "lodash": "^4.17.13",
     "moment": "^2.24.0",
     "react": "^16.4.0",
+    "react-chartjs-2": "^2.9.0",
```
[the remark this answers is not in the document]
slobodanc93: Yes, I think that we have to keep them all. Newer versions of _react-chartjs-2_ require _chart.js_ to be included as well.  _chartjs-plugin-labels_ is required because there is no out of the box solution for labels inside the graph.

src/main/frontend/src/pages/dashboard/components/styled.js
slobodanc93: I've corrected it but had to keep fixed canvas container height for smaller devices when we use row-based flex container in order for the canvas to be rendered as expected.

Issue #1302, opened 2020-04-21
Implement export in PDF format


VojislavKovacevic:
In addition to data export in CSV format, export in PDF should be implemented.  

Roles "Office Manager" and "HR Manager" should also be able to export data in PDF format.
It should be possible to export the same type/range of data as in the CSV export feature.

PR #1309, opened 2020-04-23
feature/gh-1302-implement-export-in-PDF-format

popovicbojan:
Resolves: #1302


src/main/java/eu/execom/hawaii/util/AllowancePdfExportAssembler.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Autowired
+  public AllowancePdfExportAssembler(PdfGenerator pdfGenerator) {
+    this.pdfGenerator = pdfGenerator;
+  }
+
+  public ByteArrayInputStream getAllowanceExportPdf(List<Allowance> allowances) {
+    PdfPTable table =
+        pdfGenerator.createPdfTable(
+            12, HEADER_VALUES, 100, new int[] {3, 3, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2});
```
stanislavgrujic, line 41: Please use constants with a meaningful name, instead of 12 and 100. 
Also, `new int[] { ... }` can be extracted to a constant with a meanigful name. I suspect it is column widths, but I would prefer not to have to guess.

src/main/java/eu/execom/hawaii/util/LeaveProfilesPdfExportAssembler.java
```diff
@@ -0,0 +1,55 @@
...
+  private PdfGenerator pdfGenerator;
+
+  @Autowired
+  public LeaveProfilesPdfExportAssembler(PdfGenerator pdfGenerator) {
+    this.pdfGenerator = pdfGenerator;
+  }
+
+  public ByteArrayInputStream getLeaveProfileExportPdf(List<LeaveProfile> leaveProfiles) {
+    PdfPTable table =
+        pdfGenerator.createPdfTable(9, HEADER_VALUES, 100, new int[] {3, 5, 2, 2, 2, 2, 2, 2, 2});
```
stanislavgrujic, line 36: See my comment above.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+import java.io.IOException;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+@Slf4j
+@Component
+public class PdfGenerator {
+
+  public static final String REGULAR_FONT_PATH = "src/main/resources/static/fonts/Montserrat-Regular.ttf";
+  public static final String BOLD_FONT_PATH = "src/main/resources/static/fonts/Montserrat-Black.ttf";
```
stanislavgrujic, line 32: Please make these constants private, they are not used outside of the class.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  public PdfPTable setCellStyle(PdfPCell[] cells, PdfPTable table, int index) {
+    for (PdfPCell cell : cells) {
+      cell.setVerticalAlignment(Element.ALIGN_MIDDLE);
+      cell.setHorizontalAlignment(Element.ALIGN_LEFT);
+      cell.setPadding(5);
+      cell.setBorderWidthLeft(0);
+      cell.setBorderWidthRight(0);
+      cell.setBorderWidthTop(0);
```
stanislavgrujic, line 83: These lines and lines in lines 65 to 71 are pretty similar, just in different order. Please reorder them, so that they are arranged in the same  order and extract to a method which would apply it (e.g., `applyCellFormatting(cell))`)

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+      document.add(table);
+    } catch (IOException | DocumentException e) {
+      e.printStackTrace();
+    } finally {
+      document.close();
+    }
+    return new ByteArrayInputStream(byteArrayOutputStream.toByteArray());
+  }
+
+  public PdfPCell getTableCell(String cellValue) {
```
stanislavgrujic, line 154: Please rename to `createCell(String value)`. Also, apply the cell styling immediately, there is no need to call `setCellStyle` afterwards and additionally loop through all created cells.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+    Font cellFont = null;
+    try {
+      cellFont = new Font(getBaseFont(), 8f);
+    } catch (IOException | DocumentException e) {
+      log.error("Error occurred {}", e.getMessage());
+    }
+    return new PdfPCell(new Phrase(cellValue, cellFont));
+  }
+
+  public String formatDateForPdf(LocalDate localDate) {
```
stanislavgrujic, line 164: Instead of making this method public and calling it from outside of this class, make it private, and add an extra method `createDateCell(String value)` which would format and add formatted value to the cell.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+    table.setWidthPercentage(columnsWidthPercentage);
+    try {
+      table.setWidths(columnsWidths);
+    } catch (DocumentException e) {
+      log.error("Error occurred {}", e.getMessage());
+    }
+    return setHeaderValues(headerValues, table);
+  }
+
+  private BaseFont getBaseFont() throws IOException, DocumentException {
```
stanislavgrujic, line 46: It is a good practice to put methods which are being called below the last method which calls them. E.g.,
```
method1CallsGetBaseFont() {
}

method2CallsGetBaseFont() {
}

private getBaseFont() {}
```

This makes the code easier to read and process cognitively.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+
+    Image execomLogo =
+        Image.getInstance(
+            UserPdfExportAssembler.class.getResource("/static/icons/execom_logo_red.png"));
+    header.addCell(execomLogo);
+    return header;
+  }
+
+  public ByteArrayInputStream getPdfFile(String titleValue, PdfPTable table, boolean isLandscape) {
+    ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
stanislavgrujic, line 126: Try to avoid giving technical names to your classes/methods/variables. What does this stream represent? I have no idea what it does, nor how it is used, just from reading its name.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,167 @@
...
+      cell.setBorderWidthTop(0);
+      if (index % 2 == 1) {
+        cell.setBackgroundColor(BaseColor.LIGHT_GRAY);
+      }
+      table.addCell(cell);
+    }
+    return table;
+  }
+
+  private PdfPTable getHeader(boolean isLandscape) throws IOException, DocumentException {
```
stanislavgrujic, line 92: Please rename all methods where you create something in this class from get_ to create_

src/main/java/eu/execom/hawaii/util/ResponseEntityExportAssembler.java
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ResponseEntityExportAssembler {
+
+  public <T> ResponseEntity getResponseEntity(String fileName, T bodyData, String fileFormat) {
+        var headers = new HttpHeaders();
+        headers.add("Content-Disposition", "inline; filename=\"" + fileName + "\"");
+        return ResponseEntity
```
stanislavgrujic, line 14: This indentation does not look right. Please reformat the file.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,172 @@
...
+      log.error("Error occurred {}", e.getMessage());
+    }
+    setHeaderValues(headerValues, table);
+    return table;
+  }
+
+  private void setHeaderValues(String[] headerValues, PdfPTable table) {
+    Font font = null;
+    try {
+      font = new Font(creatBoldFont(), 8f);
```
stanislavgrujic, line 50: Good approach here would be to have a FontFactory which is responsible for all types creation. You could even have static instances which you could just return here (and wrap exception inside the FontFactory method)..
```
font = FontFactory.getBoldFont(8f);
```

I assume you don't have too many fonts, so you might even create all of them and just return when needed.

src/main/frontend/src/pages/administration/components/common/DrawerElements.js
```diff
@@ -45,15 +45,24 @@ export const DrawerFooter = props => {
...
+          handleChange={handleChange}
+          value={exportFormat}
+          options={
+            <Fragment>
+              <option value="csv">CSV</option>
+              <option value="pdf">PDF</option>
+            </Fragment>
+          }
+        />
+
```
aberonja, line 65: Remove \n

src/main/frontend/src/pages/administration/components/common/DrawerElements.js
```diff
@@ -273,6 +282,22 @@ export const InputPair = props => {
   );
 };
 
+export const InputPairNoCaption = props => {
+  const { view } = props;
```
aberonja, line 286: \n

aberonja: Also, you could destructure directly on the 285 line.

src/main/frontend/src/store/sagas/helpers/sagaHelpers.js
```diff
@@ -30,7 +30,10 @@ export const extractExportDataFromResponse = response => {
   return { data, filename };
 };
 
-export const downloadCsv = (data, filename) => {
-  const csvData = new Blob([data], { type: 'text/csv;charset=utf-8;' });
-  FileSaver.saveAs(csvData, filename);
+export const downloadFile = (data, filename, exportFormat) => {
+  let formattedData =
```
aberonja, line 34: This should still be a const.

src/main/frontend/src/styles/globalStyles.js
```diff
@@ -432,6 +432,10 @@ div.drawer {
   }
 }
 
+
```
aberonja, line 435: You have \n, remove one.

src/main/frontend/src/pages/administration/components/common/DrawerElements.js
```diff
@@ -273,6 +282,22 @@ export const InputPair = props => {
...
+  return (
+    <div className="mr-3 position-relative fixed-select-width">
```
aberonja, line 288: Use w-100 instead of custom class.

src/main/java/eu/execom/hawaii/util/RequestPdfExportAssembler.java
```diff
@@ -0,0 +1,70 @@
...
+  @Autowired
+  public RequestPdfExportAssembler(PdfGenerator pdfGenerator, AllowanceService allowanceService) {
+    this.pdfGenerator = pdfGenerator;
+    this.allowanceService = allowanceService;
+  }
+
+  public ByteArrayInputStream getRequestExportPdf(List<Request> requests) {
+    int numberOfTableColumns = 12;
+    int columnsWidthPercentage = 100;
+    int[] columnsRelativeWidth = {4, 3, 4, 3, 3, 3, 3, 2, 2, 2, 2, 5};
```
stanislavgrujic, line 44: Please extract these as constants.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,172 @@
...
+        Image.getInstance(
+            UserPdfExportAssembler.class.getResource("/static/icons/execom_logo_red.png"));
+    header.addCell(execomLogo);
+    return header;
+  }
+
+  public ByteArrayInputStream createPdfFile(String titleValue, PdfPTable table, boolean isLandscape) {
+    ByteArrayOutputStream pdfFileOutputStream = new ByteArrayOutputStream();
+    int a4PaperLongSide = 842;
+    int a4PaperShortSide = 595;
```
stanislavgrujic, line 98: Please extract these as constants.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,172 @@
...
+    header.setLockedWidth(true);
+    header.getDefaultCell().setFixedHeight(40);
+    header.getDefaultCell().setBorder(Rectangle.BOTTOM);
+    header.getDefaultCell().setBorderColor(BaseColor.LIGHT_GRAY);
+
+    PdfPCell text = new PdfPCell();
+    text.setPaddingBottom(10);
+    text.setPaddingLeft(10);
+    text.setBorder(Rectangle.BOTTOM);
+    text.setBorderColor(BaseColor.LIGHT_GRAY);
```
stanislavgrujic, line 77: Extract this to a method `createTextCell`

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,172 @@
...
+
+    String[] headerItems = {
+      "Execom", "Novi Sad", "Bulevar vojvode Stepe 50", "https://www.execom.eu/"
+    };
```
stanislavgrujic, line 81: This should be a constant.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,172 @@
...
+import java.io.IOException;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+@Slf4j
+@Component
+public class PdfGenerator {
+
+  private static final String REGULAR_FONT_PATH = "src/main/resources/static/fonts/Montserrat-Regular.ttf";
+  private static final String BOLD_FONT_PATH = "src/main/resources/static/fonts/Montserrat-Black.ttf";
```
stanislavgrujic, line 32: It can be dangerous to hardcode path to your resources. Please investigate if you can use `ClassPathResource` class from Spring to load resources.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,172 @@
...
+  private PdfPTable createHeader(boolean isLandscape) throws IOException, DocumentException {
+    Font font = new Font(createBaseFont(), 6f);
+    int headerWidth = isLandscape ? 780 : 527;
+    PdfPTable header = new PdfPTable(2);
+    header.setWidths(new int[] {24, 6});
+    header.setTotalWidth(headerWidth);
+    header.setLockedWidth(true);
+    header.getDefaultCell().setFixedHeight(40);
+    header.getDefaultCell().setBorder(Rectangle.BOTTOM);
+    header.getDefaultCell().setBorderColor(BaseColor.LIGHT_GRAY);
```
stanislavgrujic, line 71: Please extract to a method `createHeader`.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,172 @@
...
+      PdfPCell hCell = new PdfPCell(new Phrase(headerValue, font));
+      applyCellFormatting(hCell);
+      hCell.setBackgroundColor(BaseColor.GRAY);
+      table.addCell(hCell);
+    }
+  }
+
+  private PdfPTable createHeader(boolean isLandscape) throws IOException, DocumentException {
+    Font font = new Font(createBaseFont(), 6f);
+    int headerWidth = isLandscape ? 780 : 527;
```
stanislavgrujic, line 64: Please extract to constants with a meaningful name.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
```diff
@@ -0,0 +1,172 @@
...
+    };
+    for (String headerItem : headerItems) {
+      text.addElement(new Phrase(headerItem, font));
+    }
+    text.setHorizontalAlignment(Element.ALIGN_LEFT);
+    header.addCell(text);
+
+    Image execomLogo =
+        Image.getInstance(
+            UserPdfExportAssembler.class.getResource("/static/icons/execom_logo_red.png"));
```
stanislavgrujic, line 90: This should be a constant, and not reside within Assembler class. Why do you use another class to load resource in this class? I suggest to create a class LogoProvider and use ClassPathResource from Spring to load logo, and use here as `header.addCell(LogoProvider.getLogo());`.

src/main/java/eu/execom/hawaii/util/UserPdfExportAssembler.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Autowired
+  public UserPdfExportAssembler(PdfGenerator pdfGenerator) {
+    this.pdfGenerator = pdfGenerator;
+  }
+
+  public ByteArrayInputStream getUserExportPdf(List<User> users) {
+    int numberOfTableColumns = 5;
+    int columnsWidthPercentage = 100;
+    int[] columnsRelativeWidth = {3, 2, 3, 2, 2};
```
stanislavgrujic, line 27: Extract constants.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
popovicbojan: This is already a part of createHeader method. I'm not sure if we should extract it.

stanislavgrujic: My bad, leave it as is.

src/main/java/eu/execom/hawaii/util/PdfGenerator.java
popovicbojan: Is it ok if I use [FontFactory](https://api.itextpdf.com/iText5/java/5.5.11/com/itextpdf/text/FontFactory.html)?
```
String regularFontPath = "static/fonts/Montserrat-Regular.ttf";
return FontFactory.getFont(
        regularFontPath,
        BaseFont.IDENTITY_H,
        BaseFont.EMBEDDED,
        fontSize,
        Font.NORMAL,
        BaseColor.BLACK);
```

stanislavgrujic: Sure, but I would wrap it inside our FontFactory, in order to hide anything which is not necessary from its users and to be able to control it in a single space.

popovicbojan: Great, that's what I've done. :)

src/main/java/eu/execom/hawaii/util/export/PdfGenerator.java
```diff
@@ -0,0 +1,161 @@
...
+  private void applyCellFormatting(PdfPCell cell) {
+    cell.setVerticalAlignment(Element.ALIGN_MIDDLE);
+    cell.setHorizontalAlignment(Element.ALIGN_LEFT);
+    cell.setPadding(5);
+    cell.setBorderWidthLeft(0);
+    cell.setBorderWidthRight(0);
+    cell.setBorderWidthTop(0);
+  }
+
+  public PdfPCell createDateCell(LocalDate date, int index) {
```
stanislavgrujic, line 153: Please apply method ordering in this class as I have previously explained. Check all classes which you have added in this PR.

src/main/java/eu/execom/hawaii/util/export/RequestPdfExportAssembler.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.stereotype.Component;
+
+import java.io.ByteArrayInputStream;
+import java.util.List;
+
+@Component
+public class RequestPdfExportAssembler {
+
+  private final int NUMBER_OF_TABLE_COLUMNS = 12;
+  private final int[] COLUMNS_RELATIVE_WIDTH = {4, 3, 4, 3, 3, 3, 3, 2, 2, 2, 2, 5};
```
stanislavgrujic, line 18: Please add `static` keyword.

src/main/java/eu/execom/hawaii/util/export/TableHeaderValues.java
```diff
@@ -0,0 +1,65 @@
+package eu.execom.hawaii.util.export;
+
+final class TableHeaderValues {
```
stanislavgrujic, line 3: Please rename to `TableHeaders` it is less verbose and still gives enough information about content of the class.

src/main/java/eu/execom/hawaii/util/export/TableHeaderValues.java
```diff
@@ -0,0 +1,65 @@
...
+
+  static final String[] ALLOWANCE_HEADER_VALUES = {
```
stanislavgrujic, line 5: Please rename to `ALLOWANCE` -> it is clear that it is values, since it is a constant, and in the class name you already have information that it is used as a table header definition. Apply the same logic for all values in this class.

src/main/java/eu/execom/hawaii/util/export/TableHeaderValues.java
```diff
@@ -0,0 +1,65 @@
+package eu.execom.hawaii.util.export;
+
+final class TableHeaderValues {
+
```
stanislavgrujic, line 4: Utility classes should also have a private constructor to make sure that noone tries to instantiate it. Please add one.

src/main/java/eu/execom/hawaii/util/export/TableHeaderValues.java
```diff
@@ -0,0 +1,65 @@
...
+    "Days",
+    "Days in Range",
+    "Reason"
+  };
+
+  static final String[] USER_PDF_HEADER_VALUES = {
+    "Name", "Team", "E-mail", "Status", "Started working"
+  };
+
+  static final String[] USER_CSV_HEADER_VALUES = {
```
stanislavgrujic, line 51: Why do we have different layout for CSV and PDF for users?

src/main/java/eu/execom/hawaii/util/export/AllowanceCsvExportAssembler.java
```diff
@@ -11,23 +11,8 @@
...
-    "Total Training",
-    "Taken Training",
-    "Pending Training",
-    "Remaining Training",
-    "Total Taken Sickness"
-  };
-
   public String getAllowanceExportCsv(List<Allowance> allowances) {
-    List<String[]> dataLines = new ArrayList<>(Collections.singletonList(HEADER));
+    List<String[]> dataLines = new ArrayList<>(Collections.singletonList(TableHeaderValues.ALLOWANCE_HEADER_VALUES));
```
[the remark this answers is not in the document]
popovicbojan: If I use that, it displays an error `Array initializer is not allowed here`. I could write it this way: `Arrays.asList(new String[][] {TableHeaderValues.ALLOWANCE_HEADER_VALUES})`, but then we create an extra array. Maybe we could write it like this:
```   
List<String[]> dataLines = new ArrayList<>();
dataLines.add(TableHeaderValues.ALLOWANCE_HEADER_VALUES);
```

src/main/java/eu/execom/hawaii/util/export/TableHeaderValues.java
popovicbojan: Users, among others, have 3 date attributes, startedWorkingDate, startedWorkingAtExecomDate, and stoppedWorkingAtExecomDate. Since pdf is more of an overview of the data I didn't want to clutter it with all these dates because most employees startedWorkingDate and startedWorkingAtExecomDate are the same and stoppedWorkingAtExecomDate is null, so I've put startedWorkingAtExecomDate in pdf report while all employees attributes are present in CSV report. Do you think I should display all attributes in the user pdf report?

src/main/frontend/src/pages/administration/components/common/DrawerElements.js
slobodanc93: In all other methods in this class, there is no \n before return... Should I omit it here for the sake of consistency?

src/main/frontend/src/pages/administration/components/common/DrawerElements.js
slobodanc93: This is not a regular input field that we have throughout the application, even w-25 is too wide on larger screens because it should look more like a button. This is the desired layout that I accomplished with that fixed-width custom class:
Capture

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -103,33 +133,124 @@ private String getAbsenceName(AbsenceType absenceType) {
...
+      LocalDate startDate,
+      LocalDate endDate) {
+    List<UserStatusType> userStatusTypes = getUserStatusTypes(includeInactiveEmployees);
+    boolean absenceTypesEmpty = absenceType == null;
+    List<AbsenceType> absenceTypes = absenceTypesEmpty ? null : setAbsenceTypes(absenceType);
+
+    return requestRepository.findAllRequestsForExport(userStatusTypes, absenceTypes, teamId, startDate, endDate, absenceTypesEmpty);
+  }
+
+  private List<AbsenceType> setAbsenceTypes(AbsenceType absenceType) {
```
stanislavgrujic, line 173: Please pay attention to method naming -> you are not setting anything here, you are retrieving a list of absence type for which you are searching. Also, it seems to me that the code is more complex than needed. I propose to use the following:
```
List<AbsenceType> absenceTypes = getAbsenceTypesForExport(absenceType);
private List<AbscenceType> getAbsenceTypesForExport(AbsenceType absenceType) {
  if (absenceType == null) {
    return null;
  }

  List<AbsenceType> absenceTypes = new ArrayList<>(List.of(absenceType));
  if (absenceType.equals(AbsenceType.LEAVE)) {
    absenceTypes.add(AbsenceType.TRAINING);
  }
  return absenceTypes;
}
```

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -103,33 +133,124 @@ private String getAbsenceName(AbsenceType absenceType) {
...
+    List<AbsenceType> absenceTypes = new ArrayList<>(List.of(absenceType));
+    if (absenceType.equals(AbsenceType.LEAVE)) {
+      absenceTypes.add(AbsenceType.TRAINING);
+    }
+
+    return absenceTypes;
+  }
+
+  public ExportFile exportUsers(
+      String exportFormat, boolean includeInactiveEmployees, Long teamId) {
```
stanislavgrujic, line 183: Why break lines? Do you have set line length of 120 characters? Please don't break lines unless needed.

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -103,33 +133,124 @@ private String getAbsenceName(AbsenceType absenceType) {
...
+
+  private String createUserExportFileName(Long teamId, boolean includeInactiveEmployees, String exportFormat) {
     String teamName = assignTeamName(teamId);
     String inactiveEmployees = assignNonActiveUsersSuffix(includeInactiveEmployees);
 
-    return "export-for-" + teamName + inactiveEmployees + FILE_SUFFIX;
+    return "export-for-" + teamName + inactiveEmployees + assignFileSuffix(exportFormat);
+  }
+
+  private <T> T getUserExportFile(String fileFormat, Long teamId, boolean includeInactiveEmployees) {
```
stanislavgrujic, line 197: Please try to use the same ordering of parameters whenever possible.
Method `exportUsers(String exportFormat, boolean includeInactiveEmployees, Long teamId)` and this one should have the same ordering. This makes it easier to read and understand code by associating parameters also by position, not just name.

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -103,33 +133,124 @@ private String getAbsenceName(AbsenceType absenceType) {
...
     return users.stream().filter(User::hasAnyRoleOtherThanHrAssistant).collect(Collectors.toList());
   }
 
-  public String getAllowanceExportFilename(int year, Long teamId, boolean includeInactiveEmployees) {
+  private String assignFileSuffix(String exportFormat) {
+    return "." + exportFormat;
+  }
+
+  public ExportFile exportAllowances(
+      int year, String exportFormat, boolean includeInactiveEmployees, Long teamId) {
```
stanislavgrujic, line 217: Please do not break lines when not needed.

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -103,33 +133,124 @@ private String getAbsenceName(AbsenceType absenceType) {
...
+    String fileFormat = determineFileFormat(exportFormat);
+    String fileName = createAllowanceExportFilename(year, teamId, includeInactiveEmployees, exportFormat);
+    var bodyData = getAllowanceExportFile(fileFormat, year, teamId, includeInactiveEmployees);
+    return new ExportFile(bodyData, fileName, fileFormat);
+  }
+
+  private String createAllowanceExportFilename(
+      int year, Long teamId, boolean includeInactiveEmployees, String exportFormat) {
```
stanislavgrujic, line 225: Please do not break lines when not needed.

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -103,33 +133,124 @@ private String getAbsenceName(AbsenceType absenceType) {
...
+    return team.map(value -> "team-" + value.getName().toLowerCase()).orElse("all-employees");
+  }
+
+  private String assignNonActiveUsersSuffix(boolean includeInactiveEmployees) {
+    return !includeInactiveEmployees ? "" : "-non-active-users-included";
   }
 
-  public List<Allowance> getAllowancesForExport(int year, Long teamId, boolean includeInactiveEmployees) {
+  private <T> T getAllowanceExportFile(
+      String fileFormat, int year, Long teamId, boolean includeInactiveEmployees) {
```
stanislavgrujic, line 245: Please do not break lines when not needed.

src/main/java/eu/execom/hawaii/util/export/AllowanceCsvExportAssembler.java
```diff
@@ -11,23 +11,9 @@
...
-    "Total Training",
-    "Taken Training",
-    "Pending Training",
-    "Remaining Training",
-    "Total Taken Sickness"
-  };
-
   public String getAllowanceExportCsv(List<Allowance> allowances) {
-    List<String[]> dataLines = new ArrayList<>(Collections.singletonList(HEADER));
+    //TODO Resolve this better
```
stanislavgrujic, line 15: When will this TODO be resolved? Leaving a TODO in code like this is a bad practice which results in hanging TODOs in code forever. Either create a story to resolve the TODO, or do it right away.

src/main/java/eu/execom/hawaii/util/export/LeaveProfilesPdfExportAssembler.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Autowired
+  public LeaveProfilesPdfExportAssembler(PdfGenerator pdfGenerator) {
+    this.pdfGenerator = pdfGenerator;
+  }
+
+  public ByteArrayInputStream getLeaveProfileExportPdf(List<LeaveProfile> leaveProfiles) {
+    PdfPTable table =
+        pdfGenerator.createPdfTable(
+            NUMBER_OF_TABLE_COLUMNS, TableHeaders.LEAVE, COLUMNS_RELATIVE_WIDTH);
```
stanislavgrujic, line 27: The line 25 would be exactly 120 charactes. Please do not break lines unless absolutely necessary.

src/main/java/eu/execom/hawaii/util/export/LogoProvider.java
```diff
@@ -0,0 +1,28 @@
...
+@Lazy
+@Component
+public class LogoProvider {
+
+  Image getLogo() {
+    Image execomLogo = null;
+    try {
+      execomLogo =
+          Image.getInstance(
+              ClassPathResource.class.getResource("/static/icons/execom_logo_red.png"));
```
stanislavgrujic, line 22: Please do not break lines -> these 3 lines should be a single one.

src/main/java/eu/execom/hawaii/util/export/PdfGenerator.java
```diff
@@ -0,0 +1,162 @@
...
+  private static final int[] HEADER_COLUMNS_WIDTH = {24, 6};
+  private static final String[] HEADER_ITEMS = {
+    "Execom", "Novi Sad", "Bulevar vojvode Stepe 50", "https://www.execom.eu/"
+  };
+
+  private CustomFontFactory customFontFactory;
+  private LogoProvider logoProvider;
+
+  @Autowired
+  @Lazy
```
stanislavgrujic, line 43: Please reverse the order of these annotations. Next to the constructor the most important (@Autowired) and then the others -> @Lazy. Similar to the way we do it with the class.

src/main/java/eu/execom/hawaii/util/export/TableHeaders.java
```diff
@@ -0,0 +1,68 @@
...
+    "Days",
+    "Days in Range",
+    "Reason"
+  };
+
+  static final String[] USER_PDF = {
+    "Name", "Team", "E-mail", "Status", "Started working"
+  };
+
+  //TODO check if user csv and pdf headers neew to be the same
```
stanislavgrujic, line 53: When will this happen?

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -109,32 +119,111 @@ private String getAbsenceName(AbsenceType absenceType) {
...
+      LocalDate startDate, LocalDate endDate) {
+    List<UserStatusType> userStatusTypes = assignUserStatusTypes(includeInactiveEmployees);
+    boolean absenceTypesEmpty = absenceType == null;
+    List<AbsenceType> absenceTypes = assignAbsenceTypesForExport(absenceType);
+
+    return requestRepository.findAllRequestsForExport(userStatusTypes, absenceTypes, teamId, startDate, endDate,
+        absenceTypesEmpty);
+  }
+
+  private List<AbsenceType> assignAbsenceTypesForExport(AbsenceType absenceType) {
```
stanislavgrujic, line 145: To what do you assign absence types? Please pay attention to method naming and usage of verbs and their meaning.

popovicbojan: Ok, so should I use `create` here instead of `assign`, because new List of absence types is created, or to revert it to `get`. I thought that `get` is not appropriate because I'm not getting data from the database or from some object but actually creating a new `List`. I've put assign because sometimes method returns null, so we don't create each time new `List`.

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -109,32 +119,111 @@ private String getAbsenceName(AbsenceType absenceType) {
...
+    return (T) new InputStreamResource(allowancePdfExportAssembler.getAllowanceExportPdf(allowances));
+  }
+
+  private List<Allowance> getAllowancesForExport(Long teamId, boolean includeInactiveEmployees, int year) {
+    List<UserStatusType> userStatusTypes = assignUserStatusTypes(includeInactiveEmployees);
     return allowanceRepository.findAllByYearTeamIdAndUserStatusType(year, teamId, userStatusTypes);
   }
 
-  private List<UserStatusType> getUserStatusTypes(boolean includeInactiveEmployees) {
+  private List<UserStatusType> assignUserStatusTypes(boolean includeInactiveEmployees) {
```
stanislavgrujic, line 226: When you say assign it implicitly tells that you are providing something to which you will add something else. In this case it is not as such. Please use another verb instead of assign.

popovicbojan: Ok, now I understand. :)

src/main/frontend/src/pages/administration/components/common/DrawerElements.js
aberonja: Yes, that would be good :)

src/main/frontend/src/store/sagas/helpers/sagaHelpers.js
```diff
@@ -30,7 +30,10 @@ export const extractExportDataFromResponse = response => {
   return { data, filename };
 };
 
-export const downloadCsv = (data, filename) => {
-  const csvData = new Blob([data], { type: 'text/csv;charset=utf-8;' });
-  FileSaver.saveAs(csvData, filename);
+export const downloadFile = (data, filename, exportFormat) => {
+  const formattedData =
+    exportFormat === 'csv'
+      ? new Blob([data], { type: 'text/csv;charset=utf-8;' })
```
aberonja, line 36: Could this check if file is 'csv' be on this line?

Issue #1314, opened 2020-04-24
Allowances report with loading wrapper


slobodanc93:
Implement a loading indicator while data is being fetched.
Currently:
![Screenshot_1](https://user-images.githubusercontent.com/42570935/80213456-7c023a00-8639-11ea-8ad2-a6c3146bd38e.png)

PR #1316, opened 2020-04-24
f-enhancement/gh-1314-allowances-report-with-loadin-wrapper

popovicbojan:
Resolves: #1314


src/main/frontend/src/store/reducers/allowanceReducer.js
```diff
@@ -11,7 +11,7 @@ import { makeUpdaterReducer } from './helpers/reducerHelpers';
 
 const initialState = {
   allowances: null,
-  results: [],
+  results: null,
```
aberonja, line 14: Is there a reason why is this null?

popovicbojan: Results is passed to `AdministrationTable` as body prop. When body prop is null (data are not fetched yet), `AdminstrationTable` renders `TablePlaceholder` component. 

image

Issue #1311, opened 2020-04-24
User can delete himself 


popovicbojan:
User with HR_MANAGER role can edit his information and delete himself from Administration/Employees tab.

PR #1325, opened 2020-04-29
f-fix/gh-1311-User-can-delete-himself

slobodanc93:
Resolves: #1311


src/main/frontend/src/pages/administration/components/employees/EmployeeInfo.js
```diff
@@ -155,11 +155,15 @@ class EmployeeInfo extends Component {
   };
 
   handleDelete = employee => {
+    const message =
+      this.isOnlyHrManager() && employee.userRole === 'HR_MANAGER'
+        ? 'This action will delete the only HR manager, no one with admin privileges will have access to the application. Are you sure you want to continue?'
+        : 'This action will delete the selected employee who will lose access to the application. Are you sure you want to continue?';
```
aberonja, line 161: This looks a bit messy, maybe it would be a good thing to extract both messages to variables somewhere outside this component. It's not a big issue, just a preference.

src/main/frontend/src/pages/administration/components/employees/EmployeeInfo.js
```diff
@@ -176,6 +180,14 @@ class EmployeeInfo extends Component {
     });
   };
 
+  isOnlyHrManager = () => {
```
aberonja, line 183: Maybe this should be called isTheOnlyHrManager(), just to be precise.

src/main/frontend/src/store/sagas/HOC/withErrorHandling.js
```diff
@@ -9,7 +9,7 @@ export function withErrorHandling(saga, errorHandler) {
       // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*
       yield* saga(...args);
     } catch (e) {
-      if (e.status === 401) {
+      if (e.status === 401 || e.status === 403) {
```
aberonja, line 12: What if user is allowed to go somewhere else except on that route?

slobodanc93:
@stanislavgrujic  When the user is deleted, the backend is responding with 403 which is currently not providing logout on the frontend (user can still log in but will see no content aside from pop up error message). The proposition is to change the response for this situation to 401 which will logout the user from the application once he/she is deleted. Please review if it is the right approach.

src/main/frontend/src/pages/administration/components/employees/EmployeeInfo.js
slobodanc93: The messages are extracted to a helper function.

src/main/frontend/src/store/sagas/HOC/withErrorHandling.js
slobodanc93: I agree that could be an issue. I'll ask for a new review once we align on this situation from the backend perspective.

src/main/frontend/src/pages/administration/components/employees/helpers.js
```diff
@@ -43,3 +43,10 @@ export const addEmailDomain = email => email.concat('@execom.eu');
 
 export const removeEmailDomain = email =>
   email.slice(0, email.lastIndexOf('@'));
+
+export const constructDeleteEmployeeMessage = isTheOnlyHrManager => {
+  const message = isTheOnlyHrManager
+    ? 'This action will delete the only HR manager, no one with admin privileges will have access to the application.'
+    : 'This action will delete the selected employee who will lose access to the application.';
+  return message.concat(' Are you sure you want to continue?');
```
aberonja, line 51: I would use `${message} Are you sure you want to continue?` here.

Issue #1331, opened 2020-05-04
Zero on dashboard graph during loading


slobodanc93:
Zero is not shown on the graph, but during graph loading (1-2 sec) zero appears when there are datasets without all values. This should be changed so that zero values never appear.

PR #1334, opened 2020-05-04
f-fix/gh-1331-zero-on-dashboard-graph-during-loading

slobodanc93:
Resolves: #1331 
The new library for labels on the graph is more popular (https://www.npmtrends.com/chartjs-plugin-datalabels-vs-chartjs-plugin-labels) and brings more flexibility around conditional rendering.


src/main/frontend/src/components/allowance-chart/helpers.js
```diff
@@ -14,15 +14,19 @@ const datasetLabel = 'Allowance';
...
+const datalabels = {
+  font: {
+    size: 14,
+    family: defaultFontFamily,
+    weight: 'bold'
+  },
+  color: defaultFontColor,
+  display: function(context) {
+    const index = context.dataIndex;
+    const value = context.dataset.data[index];
```
aberonja, line 26: \n

Issue #1329, opened 2020-04-30
Error during user role change


VojislavKovacevic:
On the frontend, it is possible to change a role to "user" for user who is "approver". (In a test, the user which is approver, with current role "office-manager" is changed to role "user")

This change will not have an effect on the backend and database.
`2020-04-30 12:29:34 INFO  - It is not allowed to set 'User' role to user who has approver teams.`

The administrator is under the false impression that the role is changed. There is no warning on the UI.

PR #1347, opened 2020-05-11
enhancement/gh-1329-error-during-user-role-change

slobodanc93:
Resolves: #1329 
With these changes, **APPROVER** is no longer part of User roles enumeration (we are isolating this attribute from user roles because it was making problems when a user role is changed while editing employees, for details please see issue #1329). Instead, ***isApprover*** flag is used to determine routes and permissions. All business logic regarding isApprover flag is on the backend side.


src/main/java/eu/execom/hawaii/dto/UserDto.java
```diff
@@ -56,6 +57,9 @@
 
   private List<UserPushTokenDto> userPushTokens = new ArrayList<>();
 
+  @JsonProperty("isApprover")
+  private boolean isApprover;
```
stanislavgrujic, line 61: Do you have to explicitly state that the JSON property name is the same as the field name? I believe it should work without the annotation as well.

src/main/java/eu/execom/hawaii/model/User.java
```diff
@@ -89,12 +92,15 @@
...
-  }
-
-  public boolean hasRoleApprover() {
-    return UserRole.APPROVER.equals(userRole);
+  public boolean isApprover() {
+    switch (userRole) {
+      case HR_MANAGER:
+        return true;
+      case USER:
+        return approverTeams.size() > 0;
```
stanislavgrujic, line 100: Please know the API of the classes you use. More expressive (and desirable) is `return !isEmpty();`.

src/main/java/eu/execom/hawaii/security/UserPrincipal.java
```diff
@@ -27,8 +28,12 @@ private UserPrincipal(Long id, String email, Collection<? extends GrantedAuthori
   }
 
   public static UserPrincipal create(User user) {
-    List<GrantedAuthority> authorities = Collections.singletonList(
-        new SimpleGrantedAuthority(user.getUserRole().toString()));
+    String userRole = user.getUserRole().toString();
+    List<GrantedAuthority> authorities =
+        new ArrayList<>(Arrays.asList(new SimpleGrantedAuthority(userRole)));
```
stanislavgrujic, line 33: Please learn the API of classes you use - `Arrays.asList` returns a `new ArrayList()` already. There is no need to wrap it into a new ArrayList again. Also, these two lines should fit into a single one.

src/main/java/eu/execom/hawaii/security/UserPrincipal.java
```diff
@@ -27,8 +28,12 @@ private UserPrincipal(Long id, String email, Collection<? extends GrantedAuthori
...
+    if ("USER".equals(userRole) && user.getApproverTeams().size() > 0) {
```
stanislavgrujic, line 34: && `!user.getApproverTeams.isEmpty()`

src/test/java/eu/execom/hawaii/service/UserServiceTest.java
```diff
@@ -199,46 +199,6 @@ public void shouldGetSearchedUsersWithoutLoggedUserForHrManagerRole() {
...
-
-    // then
-    assertThat("Expect that '4' users passed search criteria", initialUsers.size(), is(4));
-    assertThat("Expect that '3' users passed search criteria and is in team where logged user is approver",
-        users.getContent().size(), is(3));
-    assertTrue("Expect that logged user is included", users.getContent().contains(mockApprover));
-
-    verify(userRepository).searchByStatusTypeEmailAndFullName(any(), anyString(), anyString());
-    verifyNoMoreInteractions(userRepository);
-  }
```
stanislavgrujic, line 240: Are there any tests you could have written that replace these two? We have remove the approver role, but have not remove the logic behind it.

src/main/java/eu/execom/hawaii/dto/UserDto.java
slobodanc93: Without @JsonProperty, Jackson actually serializes this field into "approver". This issue is present only when using primitive boolean type. I could use the Boolean wrapper class and remove annotation. Is this solution ok?

stanislavgrujic: In that case leave it as is.

src/main/java/eu/execom/hawaii/model/User.java
slobodanc93: During additional testing, I found out that we would need a null check here because "approverTeams" does not exist in UserDto so NullPointerException occurs during mapping to User. The root of this problem is the fact that UserDto class is heavily used in the codebase. Maybe a better approach here, if you agree, would be to create a new DTO class that extends UserDto and adds up "isApprover". This way we would send this flag only when it is actually needed by the frontend to determine permissions (currently only in "/me" endpoint).

src/main/java/eu/execom/hawaii/security/UserPrincipal.java
popovicbojan: I initially didn't wrap it in ArrayList but compiler raised exception when I tried to add a new element to the List, so I wrapped it. After I investigated this furder I found that Arrays.asList returns fixed-size List backed by the specified array, so every attempt of adding or removing new elements would raise `UnsupportedOperationException`. I refactored this to:
```
List<GrantedAuthority> authorities = new ArrayList<>();
authorities.add(new SimpleGrantedAuthority(userRole));
```

popovicbojan:
We've added a test for `isApprover` method, but we are not sure if it is necessary.

src/main/frontend/src/components/common/PageHeading.js
```diff
@@ -19,9 +19,10 @@ const PageHeadingTitle = styled.h1`
 `;
 
 class PageHeading extends Component {
-  renderSearchBar = (location, userRole) => {
+  renderSearchBar = (location, user) => {
+    const { userRole, isApprover } = user;
```
aberonja, line 23: \n

src/main/frontend/src/layout/helpers.js
```diff
@@ -6,38 +6,39 @@ import Dashboard from '../pages/dashboard/Dashboard';
...
 export const routes = {
   HR_MANAGER: adminRoutes,
-  APPROVER: approverRoutes,
   OFFICE_MANAGER: officeManagerRoutes,
   USER: userRoutes
 };
 
-export const generateRoutes = role => {
+export const generateRoutes = user => {
+  const { userRole, isApprover } = user;
```
aberonja, line 38: \n

Issue #1343, opened 2020-05-07
Data export options per user type


VojislavKovacevic:
On the export page, user should have options to include/exclude from the report:

1. active users
2. deleted users
3. inactive (scheduled) users

If none of the options is chosen, only active users will be included in the export.

PR #1372, opened 2020-05-20
f-enhancement/gh-1343-data-export-options-per-user-type

slobodanc93:
Resolves: #1343 
Currently, the user can turn off all switches for employee statuses. This is not compliant with backend service that will respond with data for only active employees if nothing is sent for this filter.
With these changes, we are implementing logic that forces switch for active employees to be turned on if others are not. 
Export per user status


src/main/frontend/src/pages/data-export/DataExport.js
```diff
@@ -93,7 +93,17 @@ class DataExport extends Component {
                   view="select-without-label"
                   caption="Report Type"
                   name="reportType"
-                  handleChange={handleChange}
+                  handleChange={e => {
+                    handleChange(e);
```
aberonja, line 97: \n

src/main/frontend/src/pages/data-export/DataExport.js
```diff
@@ -144,6 +161,13 @@ class DataExport extends Component {
                                 'includeScheduledEmployees',
                                 checked
                               );
```
aberonja, line 163: \n

src/main/frontend/src/pages/data-export/DataExport.js
```diff
@@ -156,6 +182,17 @@ class DataExport extends Component {
                           name="includeArchivedEmployees"
                           handleChange={checked => {
                             setFieldValue('includeArchivedEmployees', checked);
```
aberonja, line 184: \n

Issue #1377, opened 2020-06-01
Add admin monitoring dashboard


popovicbojan:
Admin dashboard should display:
- Memory usage
- Processor usage
- Number of requests
- Response time

The dashboard can be accessed by admins and one super admin. Super admin is added at the application startup, defined by user email address from application.properties, and can't be added later. Only super admin can assign or revoke admin rights to users.

Issue #1310, opened 2020-04-24
Add Spring Actuator


slobodanc93:
Spring Actuator will help us with application performance monitoring.

popovicbojan:
Resolved with #1377

PR #1378, opened 2020-06-01
feature/gh-1377-add-admin-monitoring-dashboard

popovicbojan:
Resolves: #1377 #1310


src/main/java/eu/execom/hawaii/model/metric/HardwareMetrics.java
```diff
@@ -0,0 +1,31 @@
...
+  private double systemCpuUsage;
+
+  @NotNull
+  private double jvmCpuUsage;
+
+  @CreationTimestamp
+  private LocalDateTime timestamp;
+
+  public void addToMemory(double memory) {
+    double memoryInMb = memory / 1048576;
```
stanislavgrujic, line 28: What does this number represent? Please avoid using magic numbers/literals and extract them to constants with a meaningful name. E.g., in this case I assume it is number of bytes in a mega byte, so I suggest to use
```
private static final int MEGA_BYTES = 1_048_576;
```
Also, please note how you can insert special character `_` to make better distinction between digits from your thousands and millions.

src/main/java/eu/execom/hawaii/repository/HardwareMetricsRepository.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+
+import java.time.LocalDateTime;
+import java.util.List;
+
+public interface HardwareMetricsRepository extends JpaRepository<HardwareMetrics, Long> {
+  List<HardwareMetrics> findAllByTimestampBetween(LocalDateTime fromTimestamp, LocalDateTime toTimestamp);
+
+  @Query("SELECT MIN(hardwareMetrics.timestamp) FROM HardwareMetrics hardwareMetrics")
```
stanislavgrujic, line 13: You don't have to write a query for this, you can do something like:
`LocalDateTime findFirstOrderByTimestamp`.

src/main/java/eu/execom/hawaii/security/UserPrincipal.java
```diff
@@ -27,8 +27,19 @@ private UserPrincipal(Long id, String email, Collection<? extends GrantedAuthori
...
+
+    switch (user.getUserAdminPermission()) {
+      case NONE:
+        break;
+      case BASIC:
+        authorities.add(new SimpleGrantedAuthority("ADMIN"));
+        break;
+      case SUPER:
+        authorities.add(new SimpleGrantedAuthority("SUPER_ADMIN"));
+        break;
```
stanislavgrujic, line 42: Please always add the default case. In this case it would be something like:
```
default:
  throw new IllegalArgumentException("Unrecognized permission: " + user.getUserAdminPermission());
```

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -63,18 +76,65 @@
...
+    this.environment = environment;
+  }
+
+  /**
+   * Initializes super admin identified by an email in application.properties.
+   */
+  @EventListener(ApplicationReadyEvent.class)
+  public void init() {
+    String[] activeProfiles = environment.getActiveProfiles();
+    if (Arrays.asList(activeProfiles).contains("test")) {
```
stanislavgrujic, line 101: What is this used for?

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -63,18 +76,65 @@
...
+      return;
+    }
+    validateSuperAdminEmailPropertyIsSet();
+    User newSuperAdmin = findByEmail(superAdminMail);
+    removeOldSuperAdminsIfExist(newSuperAdmin);
+    newSuperAdmin.setUserAdminPermission(UserAdminPermission.SUPER);
+    userRepository.save(newSuperAdmin);
+    log.info("New super admin is {}.", newSuperAdmin.getFullName());
```
stanislavgrujic, line 109: Please wrap placeholder values in single quoutes, since it makes it easier to track values. -> `log.info("New super admin is '{}'", newSuperAdmin.getFullName());`

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -63,18 +76,65 @@
...
+  }
+
+  /**
+   * Checks if super admin email is set in the application.properties.
+   */
+  private void validateSuperAdminEmailPropertyIsSet() {
```
stanislavgrujic, line 115: I would rename this method to `validateSuperAdminEmail` and also validate email validity with a regex.

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -63,18 +76,65 @@
...
+  private void removeOldSuperAdminsIfExist(User newSuperAdmin) {
+    List<User> oldSuperAdmins = userRepository.findAllByUserAdminPermission(UserAdminPermission.SUPER);
+    boolean oldSuperAdminsDontExist = oldSuperAdmins.isEmpty();
+    boolean newSuperAdminIsAlreadySet = oldSuperAdmins.size() == 1 && oldSuperAdmins.contains(newSuperAdmin);
+    if (oldSuperAdminsDontExist || newSuperAdminIsAlreadySet) {
+      return;
+    }
+    oldSuperAdmins.forEach(user -> user.setUserAdminPermission(UserAdminPermission.BASIC));
+    userRepository.saveAll(oldSuperAdmins);
+    log.warn("Old super admins are revoked!");
```
stanislavgrujic, line 137: Permissions are revoked, not users. Please update to *Old super admins permissions are revoked.*.
(Also, there is no need to use exclamation marks in log lines, is it as you are yelling. :)

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -259,6 +409,22 @@ public User create(User user, User modifiedByUser) {
...
+   *
+   * @param user new user.
+   */
+  private void checkIfNewUserIsAdmin(User user) {
+    UserAdminPermission userAdminPermission = user.getUserAdminPermission();
+    switch (userAdminPermission) {
+      case SUPER:
+      case BASIC:
+        throw new ActionNotAllowedException("Admin can't be created.");
+      default:
```
stanislavgrujic, line 423: This is dangerous, in case of another role, you will also fall down here. With enums and switch statements it is a good practice to list and handle explicitly all enums, and use the default case to make sure that if a new enum value is added an exception is thrown to force the developer to handle it consciously.

src/main/java/eu/execom/hawaii/jobs/SaveApplicationMetricsJob.java
```diff
@@ -0,0 +1,168 @@
...
+            createHttpEndpointMetricOrUpdateNumberOfRequests(sample);
+            break;
+          case "http_server_requests_seconds_max":
+            createHttpEndpointMetricOrUpdateMaxResponseTime(sample);
+            break;
+          default:
+            break;
+        }
+      }
+      hardwareMetricsService.save(hardwareMetrics);
```
stanislavgrujic, line 80: If there are more than one `metricFamilySamplesEnumeration` values, it will overwrite `metricValue`, if I get this code right, since the `metricValue` is defined outside of the scope of while loop. Is that what we want? Or am I mistaken?

src/main/java/eu/execom/hawaii/service/UserService.java
popovicbojan: When running tests I don't want to initialize super admin, there will be mock user for that. When running test, `test` profile is activated so based on this check I can omit initialization.

src/main/java/eu/execom/hawaii/jobs/SaveApplicationMetricsJob.java
popovicbojan: Each metric value is stored as a collection of samples in `MetricFamilySamples`, so cpu system usage, cpu jvm usage and memory usage are stored in different `MetricFamilySamples`. In one iteration we can't get all samples, so that's why hardwareMetrics is defined outside of the scope of while loop.
`Enumeration<Collector.MetricFamilySamples> metricFamilySamplesEnumeration = prometheusRegistry.metricFamilySamples();`
collects all FamilySamples metrics for some moment from `PrometheusMeterRegistry`.

src/main/java/eu/execom/hawaii/repository/HardwareMetricsRepository.java
popovicbojan: I tried this and different variations of it, but nothing seemed to work without the Query.

src/main/frontend/src/layout/helpers.js
```diff
@@ -28,25 +34,39 @@ const officeManagerRoutes = isApprover =>
...
+  USER: userRoutes,
+  ADMIN: adminRouts
 };
 
-export const generateRoutes = user => {
-  const { userRole, isApprover } = user;
+const adminRoute = adminRouts[0];
 
+export const generateRoutes = user => {
+  const { userRole, isApprover, userAdminPermission } = user;
```
aberonja, line 49: \n

src/main/frontend/src/pages/application-admin/FilterGroup.js
```diff
@@ -0,0 +1,268 @@
...
+        .format('YYYY-MM-DDTHH:mm:ss'),
+      toTimestamp: moment().format('YYYY-MM-DDTHH:mm:ss')
+    },
+    showTimepicker: false,
+    relativeTimePickerDefault: '06:00:00'
+  };
+
+  handleTimeRangeChange = e => {
+    const { monitoringTimeRange } = this.state;
+    const timeRange = e.target.value;
```
aberonja, line 26: \n

src/main/frontend/src/pages/application-admin/FilterGroup.js
```diff
@@ -0,0 +1,268 @@
...
+  handleClick = () => {
+    const showTimepicker = !this.state.showTimepicker;
+    this.setState({ showTimepicker });
+  };
+
+  render() {
+    const fromTime = moment(this.state.monitoringTimeRange.fromTimestamp);
+    const toTime = moment(this.state.monitoringTimeRange.toTimestamp);
+    const relativeTimePickerDefault = this.state.relativeTimePickerDefault;
+    const time = getTimeRange(this.props.oldestTimestamp);
```
aberonja, line 99: \n

src/main/frontend/src/pages/application-admin/FilterGroup.js
```diff
@@ -0,0 +1,268 @@
...
+                  onChange={this.handleDatePicker}
+                  defaultValue={fromTime.date()}
+                >
+                  {time.days.map(day => (
+                    <option key={`fromDay${day}`} value={day}>
+                      {day}
+                    </option>
+                  ))}
+                </DatetimeSelect>
+                .
```
aberonja, line 131: `.`?

src/main/frontend/src/pages/application-admin/styled.js
```diff
@@ -0,0 +1,76 @@
+import styled from 'styled-components';
+
+export const LineGraphBlock = styled.div`
+  width: 100%;
+  height: 500px;
+  border: solid 1px #d3d3d3;
+  border-radius: 15px;
+  box-shadow: 6px 6px 10px #d3d3d3;
+  margin: 20px; 10px;
```
aberonja, line 9: This line is invalid.

src/main/frontend/src/pages/application-admin/styled.js
```diff
@@ -0,0 +1,76 @@
...
+  padding: 10px;
+`;
+
+export const TableBlock = styled.div`
+  width: 100%;
+  border: solid 1px #d3d3d3;
+  border-radius: 15px;
+  box-shadow: 6px 6px 10px #d3d3d3;
+  margin: 20px; 10px;
```
aberonja, line 18: Same as above.

src/main/frontend/src/pages/application-admin/styled.js
```diff
@@ -0,0 +1,76 @@
...
+export const TimestampFilterGroup = styled.div`
+  display: flex;
+  margin-top: 6px;
+  margin-left: auto;
+  margin-right: 20px;
+  padding-right: 18px;
+  padding-left: 2px;
+  border: 1px solid darkgray;
+  border-radius: .25rem
+  z-index: 0;
```
aberonja, line 42: Do we need `z-index: 0`?

src/main/frontend/src/pages/application-admin/styled.js
```diff
@@ -0,0 +1,76 @@
...
+`;
+
+export const DatetimeSelect = styled.select`
+  border: none;
+  -webkit-appearance: none;
```
aberonja, line 47: Check if React add browser prefixes by default, or do we have some kind of a package for that.

src/main/frontend/src/pages/application-admin/styled.js
```diff
@@ -0,0 +1,76 @@
...
+    box-shadow: 0px 4px 6px 0px rgba(0, 0, 0, 0.2);
+    cursor: pointer;
+  }
+`;
+
+export const IconContainer = styled.div`
+  width: fit-content;
+  padding: 10px;
+  border-radius: 50%;
+  float: right;
```
aberonja, line 73: Float ❌

src/main/frontend/src/store/reducers/adminReducer.js
```diff
@@ -0,0 +1,50 @@
...
+const initialState = {
+  admin: null,
+  users: {
+    isFetching: false,
+    results: []
+  }
+};
+
+const actionHandlers = {
+  [recieveAdmins]: makeUpdaterReducer('admins'),
```
aberonja, line 19: *receive

src/main/frontend/src/store/reducers/adminReducer.js
```diff
@@ -0,0 +1,50 @@
...
+    return {
+      ...state,
+      isFetching: true
+    };
+  },
+  [superAdminSearchUsersSuccess]: (state, action) => {
+    const users = {
+      isFetching: false,
+      results: [...action.payload]
+    };
```
aberonja, line 30: \n

src/main/frontend/src/store/reducers/adminReducer.js
```diff
@@ -0,0 +1,50 @@
...
+    return {
+      ...state,
+      users
+    };
+  },
+  [clearUsers]: state => {
+    const users = {
+      isFetching: false,
+      results: []
+    };
```
aberonja, line 40: \n

src/main/frontend/src/styles/globalStyles.js
```diff
@@ -443,4 +444,82 @@ div.request-button {
     padding-right: 8px;
   }
 }
+
+div.table-header {
+  text-align: center;
+  fontSize: 15px;
+  fontFamily: 'Montserrat';
```
aberonja, line 451: Only this header have different font?

src/main/frontend/src/styles/globalStyles.js
```diff
@@ -443,4 +444,82 @@ div.request-button {
...
+  padding: 25px;
+  font-weight: 700;
+
+  @media (max-width: 1095px) {
+    font-size: 15px;
+  }
+}
+
+.delete-admin-icon {
+  float: right;
```
aberonja, line 462: Float ❌

src/main/frontend/src/styles/globalStyles.js
```diff
@@ -443,4 +444,82 @@ div.request-button {
...
+}
+
+select.no-border {
+  border: none;
+  margin-right: 24px;
+}
+
+div.timestamp-picker {
+  margin-left: auto;
+  margin-right: 244px;
```
aberonja, line 481: 🐟 
Is there another way?

src/main/frontend/src/styles/globalStyles.js
```diff
@@ -443,4 +444,82 @@ div.request-button {
...
+.dropdown-content {
+  padding: 10px;
+  border: solid 1px gray;
+  display: none;
+  position: absolute;
+  background-color: #f9f9f9;
+  min-width: 160px;
+  box-shadow: 0px 8px 16px 0px rgba(0,0,0,0.2);
+  z-index: 1;
+  right: -1px;
```
aberonja, line 511: Do we need all of this since it is display none and position absolute?

src/main/frontend/src/pages/application-admin/FilterGroup.js
popovicbojan: It's point after day in timestamp selector.
image

src/main/frontend/src/styles/globalStyles.js
popovicbojan: I made a mistake, Montserrat is default font, this is unnecessary.

src/main/frontend/src/styles/globalStyles.js
popovicbojan: Actually, this isn't used anywhere, I forgot to remove it. :)

src/main/frontend/src/styles/globalStyles.js
popovicbojan: When clicked on dropdown button dropdown-content is visible, display will be changed to flex.

src/main/java/eu/execom/hawaii/api/controller/MetricController.java
```diff
@@ -0,0 +1,61 @@
...
+    return new ResponseEntity<>(hardwareMetricDtos, HttpStatus.OK);
+  }
+
+  @GetMapping("/oldestTimestamp")
+  public ResponseEntity<LocalDateTime> getOldestTimestamp() {
+    LocalDateTime oldestTimestamp = hardwareMetricsService.getOldestTimestamp();
+    return new ResponseEntity<>(oldestTimestamp, HttpStatus.OK);
+  }
+
+  @GetMapping("/http")
```
stanislavgrujic, line 56: Do you use these 3 endpoints always at the same time, or you invoke them separately? If you always call them at the same time, it is better to have a single call with all the data in its response.

popovicbojan: They are called at the same time when React component loads. I'll make `/api/metrics` endpoint which will return MetricsDto with all this data.

popovicbojan: Sorry I made a mistake. Endpoints are called at the same time when the React component initially loads. But after that we can filter different time ranges for hardware monitoring data, so that endpoint is called separately.

src/main/frontend/src/pages/application-admin/AddAdminButton.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import Add from '@/img/icons/add.svg';
+import { Icon } from '@/pages/administration/components/years/styled';
+import { IconContainer } from './styled';
+
+const AddAdminButton = props => {
+  return (
+    <IconContainer
+      className="bg-danger"
+      onClick={e => {
```
aberonja, line 10: Is e necessary here?

popovicbojan: No it's not. I'll correct it now. :)

Issue #1368, opened 2020-05-15
Security Misconfiguration


jglugonja:
For POST request with URL http://localhost:8080/api/users/:id error message is shown: 

{"error":"Cannot deserialize value of type `eu.execom.hawaii.model.enumerations.UserRole` from String \"HR MANAGER\": not one of the values accepted for Enum class: [HR_MANAGER, HR_ASSISTANT, OFFICE_MANAGER, USER, APPROVER]\n at [Source: (PushbackInputStream); line: 5, column: 15] (through reference chain: eu.execom.hawaii.dto.UserWithApproverTeamsDto[\"userRole\"])"}

This error message contain sensitive information.

PR #1379, opened 2020-06-02
b-fix/gh-1368-security-misconfiguration

popovicbojan:
Resolves: #1368


src/main/java/eu/execom/hawaii/api/controller/GlobalExceptionHandler.java
```diff
@@ -152,8 +153,19 @@
...
-    return errorResponse(ex.getMostSpecificCause().getMessage(), HttpStatus.BAD_REQUEST);
+    return errorResponse(message, HttpStatus.BAD_REQUEST);
+  }
+
+  private String createInvalidFormatExceptionMessage(HttpMessageNotReadableException ex) {
+    String fullFieldName = ex.getMessage().split("`")[1];
+    String[] fullFieldNameArray = fullFieldName.split("\\.");
+    String field = fullFieldNameArray[fullFieldNameArray.length - 1];
+    String value = ex.getMessage().split("\"")[1];
+    return "Value '" + value + "' for type '" + field + "' is invalid.";
```
stanislavgrujic, line 168: Please use String.format instead, it is much more readable:
```
private static final String INVALID_FORMAT_MESSAGE = "Value '%s' for type '%s' is invalid.";
String.format(INVALID_FORMAT_MESSAGE, value, type);
``

Issue #1375, opened 2020-05-28
Error during sickness report export


VojislavKovacevic:
Error during sickness report export

scenario:
chouse sickness report; all Employees; 
(active/archived, date range and PDF/CSV format do not affect system behavior)
pres download

Result:
data are not exported
Message popup containing text "error"

Export per team, work as expected
Export of only archived employees works as expected

Log:
```
2020-05-28 13:24:09 ERROR - Requesting resource: '/api/export/requests' raised exception: '{}'
java.lang.NullPointerException: null
        at eu.execom.hawaii.util.export.RequestCsvExportAssembler.getRequestExportColumn(RequestCsvExportAssembler.java:55)
        at eu.execom.hawaii.util.export.RequestCsvExportAssembler.getRequestExportCsv(RequestCsvExportAssembler.java:33)
        at eu.execom.hawaii.service.DataExportService.getRequestExportFile(DataExportService.java:129)
        at eu.execom.hawaii.service.DataExportService.exportRequests(DataExportService.java:79)
        at eu.execom.hawaii.api.controller.DataExportController.exportRequests(DataExportController.java:41)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:190)
        at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:138)
        at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:104)
        at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:892)
        at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.handleInternal(RequestMappingHandlerAdapter.java:797)
[...]

PR #1381, opened 2020-06-02
b-fix/gh-1375-error-during-sickness-report-export

popovicbojan:
Resolves: #1375 
When importing requests, pending status wasn't handled, so pending requests were persisted with null status.


src/main/resources/db/changelog/changelog-master.xml
```diff
@@ -17,4 +17,9 @@
         <addUniqueConstraint columnNames="name"
                              tableName="leave_profile"/>
     </changeSet>
+    <changeSet author="admin" id="{initialId}-31">
+        <addNotNullConstraint columnName="request_status"
```
stanislavgrujic, line 21: Please also add `@NotNull` annotation in the Request class for this field.

Issue #1366, opened 2020-05-15
On team calendar, icon for holidays missing


VojislavKovacevic:
On team calendar, icon for holidays missing
in example 1.,2. and 7. January are missing icon
![teamCalendarHolidaysmissing2020-05-15 13_08_34-Hawaii execom](https://user-images.githubusercontent.com/63228397/82044432-a336af00-96ad-11ea-94aa-240bc03d5138.png)

PR #1382, opened 2020-06-02
f-fix/gh-1366-on-team-calendar-icon-for-holidays-missing

popovicbojan:
Resolves: #1366


src/main/frontend/src/pages/team-calendar/components/styled.js
```diff
@@ -249,3 +249,24 @@ export const ViewIcon = styled(({ isSelected, ...rest }) => (
     }
   }
 `;
+
+export const HolidayIconContainer = styled.div`
+  float: right;
```
aberonja, line 254: Lets not use float if we use flex. Try to do it without it.

aberonja: \n

src/main/frontend/src/pages/team-calendar/components/styled.js
```diff
@@ -249,3 +249,24 @@ export const ViewIcon = styled(({ isSelected, ...rest }) => (
...
+  @media (max-width: 767px) {
+    position: relative;
+    left: 17px;
+    bottom: 30px;
+    float: none;
```
aberonja, line 259: Same here.

src/main/frontend/src/pages/team-calendar/components/styled.js
```diff
@@ -249,3 +249,24 @@ export const ViewIcon = styled(({ isSelected, ...rest }) => (
...
+
+export const DayNumberContainer = styled.div`
+  display: inline-block;
+  padding: 4px 8px;
+  z-index: 1;
+`;
+
+export const CalendarViewImage = styled.img`
+  width: 20px;
+  margin-right: 1px;
```
aberonja, line 271: This is a bit odd, why do we have this?

popovicbojan: This is styled image for holiday icon in Team Calendar tab calendar view. I think it looks a little bit nicer with margin 1px than without it. :)
image

Issue #1393, opened 2020-06-04
Alert admins if system becomes slow and unresponsive 


popovicbojan:
Alerts should occur if:

- Invocation of endpoint produces error multiple times
- Response time of endpoint is higher than 1 second multiple times
- System and JVM CPU usage is higher than 80% over a period of time (configurable)
- Memory usage is higher than some threshold (configurable) over a period of time (configurable)

VojislavKovacevic:
When the response time of endpoint is higher than 1 second multiple times, warning mail is not dispatched to super admin.

In case of high usage of system resources and multiple errors, warning system work as expected.

PR #1405, opened 2020-06-09
b-enhancement/gh-1393-alert-admins-if-system-becomes-slow-and-unresponsive

popovicbojan:
Resolves: #1393


src/main/java/eu/execom/hawaii/service/HttpEndpointMetricsService.java
```diff
@@ -60,4 +77,26 @@ public HttpEndpointMetrics getHttpEndpointMetricIfExists(HttpEndpointMetrics htt
...
+  public void checkIfHttpEndpointAlarmNeedsToBeTriggered(HttpEndpointMetrics httpEndpointMetrics) {
+    LocalDateTime lastFailAlarmTimestamp = httpEndpointMetrics.getLastFailAlarmTimestamp();
+    if (lastFailAlarmTimestamp != null) {
+      LocalDateTime currentTimestamp = LocalDateTime.now();
+      long minutesSinceLastAlarm = ChronoUnit.MINUTES.between(lastFailAlarmTimestamp, currentTimestamp);
+      if (minutesSinceLastAlarm < snoozeTime) {
+        return;
+      }
+    }
+    boolean isExceptionRaised = !NO_EXCEPTION.equals(httpEndpointMetrics.getException());
```
stanislavgrujic, line 95: This should be a method in `HttpEndpointMetrics#isExceptionRaised()`. Doing it this way (with external check) you are putting outside logic which you might need in another place as well, and also it is internal logic to *know* what does *None* value mean for field `exception` (basically, you are breaking encapsulation).

src/main/java/eu/execom/hawaii/model/metric/HttpEndpointMetrics.java
```diff
@@ -25,4 +29,10 @@
 
   @NotNull
   private double maxResponseTime;
+
+  private LocalDateTime lastFailAlarmTimestamp;
+
+  public boolean isExceptionRaised() {
+    return !NO_EXCEPTION.equals(this.getException());
```
stanislavgrujic, line 36: No need to call `getExceptio()` - you should just use `this.exception`;

Issue #1406, opened 2020-06-09
Super admin removed


VojislavKovacevic:
Super admin status is conditioned by the HR_MANAGER role. If the Hr_manager role is removed for super admin, his status as super admin will be removed. Another HR manager is able to change super admin role via tab administration->employees

Status of super admin should be independent of the HR manager role or change of super admin role should be prevented.

VojislavKovacevic:
If app is restarted super admin status is reassigned

PR #1409, opened 2020-06-09
b-fix/gh-1406-super-admin-removed

popovicbojan:
Resolves: #1406 
The problem occurred when Hr Manager updated super admin from the Administration tab.  `UserDto `with updated information didn't contain `userAdminPermission `attribute so after update super admin permission level was set to the default value `NONE`. Because of that I changed `update` method in `UserService` so that it doesn't change `userAdminPermission` while updating other information. UserAdminPermission level can only be changed from `updateAdmin` method.


src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -388,6 +355,39 @@ private void validateAdminPermissionIsRevokedBySuperAdmin(User oldUser, User upd
     }
   }
 
+  private void validateAdminPermissionIsNotSuper(User oldUser, User updatedUser) {
```
stanislavgrujic, line 358: I think better naming would be `validateSuperAdminRoleChange`

src/main/java/eu/execom/hawaii/service/UserService.java
```diff
@@ -388,6 +355,39 @@ private void validateAdminPermissionIsRevokedBySuperAdmin(User oldUser, User upd
...
+    boolean isOldPermissionSuperAdmin = UserAdminPermission.SUPER.equals(oldUser.getUserAdminPermission());
+    boolean isNewPermissionSuperAdmin = UserAdminPermission.SUPER.equals(updatedUser.getUserAdminPermission());
+    if (isOldPermissionSuperAdmin && !isNewPermissionSuperAdmin) {
+      throw new ActionNotAllowedException("Super admin role can't be revoked.");
+    }
+    if (isNewPermissionSuperAdmin && !isOldPermissionSuperAdmin) {
+      throw new ActionNotAllowedException("Super admin can't be added.");
+    }
```
stanislavgrujic, line 366: You can simplify this code:
```
if (isOldPermissionSuperAdmin != isNewPermissionSuperAdmin) {
  throw new ActionNotAllowedException("Super admin role cannot be assigned or revoked.");
}
```

Issue #1394, opened 2020-06-05
Enable sorting of export data 


VojislavKovacevic:
Enable sorting of export data 

1. Users should have the ability to export request data sorted by user name/mail, team, status, date, or combination of them.
If combined data should be sorted in order 1.Name/mail 2.date 3. status
If export per team is selected, sorting by team may be disabled.

2. Users should have the ability to export user data sorted by the name, team, status if all employee export is selected. and name and status if data is exported per team.

PR #1421, opened 2020-06-15
feature/gh-1394-enable-sorting-of-export-data

popovicbojan:
Resolves: #1394


src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -122,22 +133,92 @@ private String assignDateRange(LocalDate startDate, LocalDate endDate) {
...
+    String defaultProperty = DEFAULT_SORT_PROPERTIES.get(exportType);
+    return Sort.by(new Sort.Order(defaultDirection, defaultProperty));
+  }
+
+  /**
+   * Throws exception if number of sort properties exceeds maximum allowed.
+   *
+   * @param propertiesLength number of properties for multiple sorting.
+   */
+  private void throwExceptionIfNumberOfSortPropertiesExceeded(int propertiesLength) {
```
stanislavgrujic, line 183: Please rename this method to `validateNumberOfSortProperties`.

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -122,22 +133,92 @@ private String assignDateRange(LocalDate startDate, LocalDate endDate) {
...
+    if (propertiesLength > MAX_SORT_PROPERTIES) {
+      String message = String.format(EXCEEDED_NUMBER_OF_PROPERTIES_MASSAGE, MAX_SORT_PROPERTIES);
+      log.error(message);
```
stanislavgrujic, line 186: Exception should be logged so you can skip this line.

src/main/java/eu/execom/hawaii/service/DataExportService.java
```diff
@@ -122,22 +133,92 @@ private String assignDateRange(LocalDate startDate, LocalDate endDate) {
...
+
+  /**
+   * Checks if sort direction is set for each property. If not default direction ASC will be set.
+   *
+   * @param propertiesLength number of sort properties.
+   * @param directions list of set directions.
+   */
+  private void checkIfSortDirectionIsSetForEveryProperty(int propertiesLength, List<Sort.Direction> directions) {
+    int directionsLength = directions.size();
+    if (propertiesLength > directionsLength) {
```
stanislavgrujic, line 199: Can't you skip the `if` check and just loop and add missing `Sort.Direction`s?

popovicbojan: Yes, you are right, `if` check is unnecessary.

src/main/frontend/src/pages/data-export/DataExport.js
```diff
@@ -18,16 +18,25 @@ import DrawerCalendarInput, {
...
   ALLOWANCES_REPORT,
-  LEAVE_PROFILES_REPORT
+  LEAVE_PROFILES_REPORT,
+  ASC,
+  DESC
 } from '@/helpers/enum';
 
 class DataExport extends Component {
+  state = {
+    directions: [DESC, ASC, ASC]
```
aberonja, line 37: Why do we have ASC twice?

popovicbojan: We are sorting data by three properties defined by the name of the property and sort direction. Initial sort direction values are `desc, asc, asc`.

PR #1437, opened 2020-06-24
feature/gh-1428-enable-in-app-notifications_model-and-requests-logic

popovicbojan:


src/main/java/eu/execom/hawaii/configuration/WebSocketConfig.java
```diff
@@ -0,0 +1,32 @@
...
+@EnableWebSocketMessageBroker
+public class WebSocketConfig implements WebSocketMessageBrokerConfigurer {
+  @Value("${hawaii.localhost}")
+  private String local;
+  @Value("${hawaii.localhost.dev}")
+  private String localDev;
+  @Value("${hawaii.staging}")
+  private String staging;
+  @Value("${hawaii.production}")
+  private String production;
```
stanislavgrujic, line 20: Could you explain how come we have to support all of these environments at once? I would expect that we only have one such configuration property. Also, this kind of configuration is forcing one to know configuration data (URLs) of all environments in each of those environments, which does not make much sense (why would production need to know about dev or staging environment)

src/main/java/eu/execom/hawaii/security/TokenProvider.java
```diff
@@ -78,4 +68,19 @@ boolean validateToken(String authToken) {
     }
     return false;
   }
+
+  String getUserEmailFromToken(String token) {
```
stanislavgrujic, line 72: It is implicitly clear that you are retrieving the information from token, because you have named your parameter *token*. You can remove part of the method name *_fromToken*. It will also make the method naming consistent with *getClaims()*.

src/main/java/eu/execom/hawaii/security/TokenProvider.java
```diff
@@ -78,4 +68,19 @@ boolean validateToken(String authToken) {
...
+    return getClaims(token).getSubject();
+  }
+
+  String getIssuerFromToken(String token) {
```
stanislavgrujic, line 76: See my comment above.

src/main/java/eu/execom/hawaii/service/AlertService.java
```diff
@@ -0,0 +1,172 @@
...
+  private String createRequestNotificationMessage(Request request) {
+    AbsenceType absenceType = request.getAbsence().getAbsenceType();
+    String userFullName = request.getUser().getFullName();
+    String leavePeriodMessage = createLeavePeriodMessage(request);
+    RequestStatus requestStatus = request.getRequestStatus();
+    if (requestStatus.equals(RequestStatus.PENDING)) {
+      return String.format(NEW_LEAVE_NOTIFICATION_TEMPLATE, userFullName, ABSENCE_TYPES.get(absenceType),
+          leavePeriodMessage);
+    } else {
+      if (AbsenceType.SICKNESS.equals(absenceType)) {
```
stanislavgrujic, line 152: Please be consistent in how you compare annotation values. Is it `variable.equals(TYPE_VALUE)` or `TYPE_VALUE.equals(variable)`. Do it one way or the other, but do it consistently throughout the code base. My preference would be to do it the second way because it makes it null safe.

src/main/java/eu/execom/hawaii/configuration/WebSocketConfig.java
slobodanc93: We took this logic already applied in `WebConfig` class when defining allowed origins.
I do agree that we don't have to support all environments at once.
In application.properties file we have `server.name` property which could be utilized for this purpose. Usage of this property is currently only for knowing where to redirect the user when clicking on a link in an email.

